PR #3, opened 2026-06-18
fix(wow-packet): write 9 movement sub-bits in create-block for 3.4.3.54261

jarlbrak:
## Problem

The real 3.4.3.54261 client crashes with ERROR#132 (physics out-of-bounds write) at world entry while parsing the `SMSG_UPDATE_OBJECT` create-block.

## Cause

`write_movement_update` emitted **7** conditional movement sub-bits before the speed block, with no `flush_bits`. The 54261 client reads **9** sub-bits there (the trailing `HasAdvFlying` + `HasDriveStatus`). With only 7 written, the later flush emits 1 byte instead of 2, so the 9 movement speeds land 1 byte early — the client reads garbage RunBack/FlightBack/TurnRate and crashes in the physics tick.

## Fix

Write the two trailing sub-bits (`HasAdvFlying`, `HasDriveStatus`) and `flush_bits()` so the speed block aligns where the 54261 client expects it. Verified empirically: with 9 sub-bits the client parses the create-block (decodes Walk=2.5 / Run=7.0 / RunBack=4.5 / Swim=4.722 / …) and reaches world-load; with 7 or 8 it crashes.

## C++ parity

This matches **current upstream TrinityCore wotlk_classic**. `Object::BuildMovementUpdate` (`Object.cpp`) writes **9** bits in the create-block, ending `HasInertia` → `HasAdvFlying` → `HasDriveStatus`, then `FlushBits()`. `HasDriveStatus` is not a padding constant — it gates a real, currently-maintained `MovementInfo::driveStatus` field (`if (HasDriveStatus) { WriteBit(accelerating); WriteBit(drifting); << speed; << movementAngle; }`) written right before the speeds. So writing 9 here is straight parity with the create-block path, not a client-only deviation.

Note: the standalone `MovementInfo` writer in `MovementPackets.cpp` (a different opcode path) still ends at `HasAdvFlying` (8 bits). This PR only touches the create-block path (`Object.cpp` / `SMSG_UPDATE_OBJECT`), which is the 9-bit one.

## Byte alignment

`FlushBits` pads to a byte boundary: 8 sub-bits → 1 byte, 9 sub-bits → 2 bytes. The 8↔9 boundary is a full byte, which is exactly why 7 or 8 still crashes 54261 (speeds shifted one byte early) and 9 aligns.

## Regression test

Added `movement_create_block_flushes_nine_subbits_to_two_bytes_before_speeds_like_cpp` (`wow-packet`), which pins the speed block to the +2-byte boundary after the 9 flushed sub-bits and decodes all 9 speeds in order. Red/green verified: it fails if the write is reverted to 7/8 sub-bits.

## 51943 compatibility

51943 is an earlier 3.4.3 build, likely from before the Drive system entered the client's movement parser, so it may expect 8 bits here. If so, the fully faithful long-term form is to gate the 9th (`HasDriveStatus`) bit on the client build — happy to make it build-conditional. This PR is validated against 54261, the build I can test.


alseif0x:
Thanks a lot for tracking this down and for testing it against the real 3.4.3.54261 client.

This is definitely the right area: Rust currently writes only 7 movement condition bits in the CreateObject movement block, and that can absolutely misalign the speed block and explain the client crash.

Before merging, I want to be strict about one point because this project is a C++ -> Rust port and we always contrast against the legacy C++ source:

- `MovementPackets.cpp` writes 8 bits here, ending at `HasAdvFlying`, then `FlushBits()`.
- `Object.cpp` also writes 8 bits here, ending at `HasAdvFlying`.
- I could not find `HasDriveStatus` in our legacy C++ tree.

So I think we need one of these two outcomes:

1. If the goal is exact C++ parity, the fix should be: add the missing `HasAdvFlying` bit and flush after 8 bits.
2. If the real 54261 client truly expects 9 bits, then keeping `HasDriveStatus = false` may be correct, but we should document it as an intentional client-compat deviation from the legacy C++ source, not as plain C++ parity.

Please also add a focused `wow-packet` layout test that proves the movement speed block is aligned correctly after these condition bits. This packet is fragile enough that I do not want to merge it without a regression test.

The current CI failure looks unrelated to this diff, but I would still prefer to merge once the PR is green or once we have explicitly confirmed it is a known base-branch failure.

I may also test this tomorrow with the actual client on our side, because if the client really requires the 9th bit then we should accept that and document why.

jarlbrak:
Thanks — and good call pushing on C++ parity. I dug into this, and I think the 8-vs-9 difference is two different points on the TrinityCore timeline rather than a client-only deviation:

- The legacy tree you checked writes 8 bits / no `HasDriveStatus` because it predates the Drive system landing in the create-block.
- **Current upstream** TrinityCore wotlk_classic `Object::BuildMovementUpdate` (`Object.cpp`) writes **9** bits: `… HasInertia → HasAdvFlying → HasDriveStatus`, then `FlushBits()`. And `HasDriveStatus` isn't a pad — there's a real `MovementInfo::driveStatus` field with an `if (HasDriveStatus) { WriteBit(accelerating); WriteBit(drifting); << speed; << movementAngle; }` payload block written just before the speeds. So writing 9 here is exact parity with the current create-block path (your option 1), just against a newer baseline than the legacy tree.

One nuance: the standalone `MovementInfo` writer in `MovementPackets.cpp` (a different opcode) does still end at `HasAdvFlying` (8 bits). This PR only touches the create-block path (`Object.cpp` / `SMSG_UPDATE_OBJECT`), which is the 9-bit one.

Why 8 still crashes 54261: `FlushBits` pads to a byte, so 8 bits → 1 byte but 9 bits → 2 bytes. That one-byte shift throws the 9 speeds off and trips ERROR#132 — which also means the legacy 8-bit write would itself crash a 54261 client.

Layout test: added `movement_create_block_flushes_nine_subbits_to_two_bytes_before_speeds_like_cpp` in `wow-packet`. It pins the speed block to the +2-byte boundary after the flushed sub-bits and decodes all 9 speeds in order; I red/green-verified it by reverting to 8 bits (fails) and back to 9 (passes). Pushed in 320b384.

On 51943: you're right to flag it. It's an earlier build and likely predates Drive, so it may want 8 here — if your client test tomorrow confirms that, the faithful fix is to gate the 9th bit on the client build, and I'm happy to make it build-conditional.

CI: the failing job is 2 `wow-world` `session.rs` tests (`dispatch_metadata_matches_cpp_for_registered_active_opcodes` and `represented_mounted_aura_toggles_mount_flag_like_cpp`). This is a packet-only diff that doesn't touch `wow-world`, so those are pre-existing/unrelated; the new push re-runs CI.

alseif0x:
Thanks for the detailed follow-up. I verified this against current TrinityCore `wotlk_classic` upstream:

- `Object::BuildMovementUpdate` does write `HasAdvFlying` and then `HasDriveStatus`.
- `HasDriveStatus` also has a real conditional payload block afterwards.
- Our local legacy C++ tree lacks that field, so this is a version/timeline divergence, not an arbitrary client-only workaround.

So I agree the 9-bit layout is the correct upstream-parity shape for the modern `wotlk_classic` create-block / `3.4.3.54261` path.

The remaining merge blocker for me is compatibility with the client build currently tested on this project (`51943`). The README still says target `3.4.3.54261`, but our live smoke tests have been against `51943`, so I do not want to merge a packet-layout change that may make that build regress without one of these being true:

1. the packet writer is build-aware and keeps the 8-bit legacy layout for builds that predate Drive, or
2. we explicitly decide RustyCore is now treating this path as `54261-only`, document that, and verify it with the real client.

The focused test you added is the right kind of regression test for the `54261` layout. I also reproduced the relevant tests locally in an isolated worktree and the packet/world tests passed here, so the current GitHub focused-test failure looks unrelated to this diff.

I am not merging this yet only because of the `51943` vs `54261` compatibility question. If you already know where the negotiated client build is available in this packet path, build-gating this would make the PR much safer. Otherwise I can test our client side and decide whether this should be accepted as `54261-only` with a follow-up for build-aware packet layouts.

alseif0x:
Thanks again for digging into this. We tested the failing path today with the real project client, and the result changes the conclusion for this repository.

You were right about the crash area: the failure is in the `SMSG_UPDATE_OBJECT` create-block movement layout, and the symptom was the same `ERROR #132` / world-entry crash class.

But the confirmed root cause here is the opposite of this PR's proposed fix:

- Our canonical C++ source for this port (`woltk-trinity-legacy`) writes 8 movement create sub-bits in `Object::BuildMovementUpdate`, ending at `HasAdvFlying`.
- It does not write `HasDriveStatus` in this path.
- With Rust writing the ninth `HasDriveStatus` bit, the 3.4.3.54261 client used by this project crashed on world entry.
- Forcing the C++ 8-bit layout made the same client enter the world successfully.
- I then converted that diagnostic into the production fix in `develop`: `2ba97a19 Use C++ movement create subbit layout`.

So I think the PR identified the right packet family, but the upstream/newer TrinityCore timeline assumption does not match the C++ baseline and client behavior we are using for RustyCore right now. The current conflict is therefore not just a merge conflict to resolve by rebasing; merging this as-is would reintroduce the client crash we just isolated.

For now I am going to keep the RustyCore path on the C++-verified 8-bit layout. If we later add explicit multi-build packet layout support and prove a different build really needs `HasDriveStatus`, we can handle that as a build-gated compatibility path instead of changing the default layout.

alseif0x:
I resolved the branch conflict by merging current `develop` into this PR branch and keeping the C++-verified 8-bit movement create layout from `develop`.

GitHub now reports the PR as mergeable with 0 changed files. That means the correct fix is already present on `develop` via `2ba97a19 Use C++ movement create subbit layout`, and there is no remaining patch from this PR to merge.

I am going to close this as superseded rather than merge it, because merging it would only bring the older 9-bit commits into history even though the final tree matches `develop`. The useful outcome from this PR was identifying the fragile packet area; the final fix landed with the layout verified against our canonical C++ source and the real client test.

alseif0x:
One last request: if you still have the same 3.4.3.54261 client/environment available, please test current `develop` with the 8-bit C++-verified layout from `2ba97a19`.

If it works on your side too, that confirms this as the correct RustyCore fix.

If it does not work on your side, please share the concrete evidence so we can compare environments/builds instead of guessing:

- exact client build shown by the client,
- whether the crash happens before or after character/world entry,
- the client error (`ERROR #132`, `WOW51900319`, etc.),
- the packet/layout observation if you have one,
- and ideally a minimal reproduction or packet trace.

If we can prove that different 3.4.3 builds really require different create-block movement layouts, the right follow-up would be a build-gated compatibility path rather than changing the default away from our canonical C++ baseline.

PR #4, opened 2026-06-22
fix(wow-packet): omit reserved ActivePlayerData field bit 50 from VALUES update

jarlbrak:
## Summary

ActivePlayerData field bit 50 (`ShieldBlockCritPercentage`) is reserved in the WoW 3.4.3.54261 client grammar — it has no backing property and is silently skipped by the client's update-field walker. rustycore's stats VALUES-update path (`write_active_player_data_values_update`) set mask bit 18 of `block[1]` for this field and then wrote a 4-byte float for it. This shifted every subsequent field in the ActivePlayerData values block by +4 bytes, desyncing the client's value walk shortly after entering the world.

## Changes (`crates/wow-packet/src/packets/update.rs`)

- **Mask:** `blocks[1] |= 0xFFFF_FFC0` → `0xFFFB_FFC0` (clears bit 18 of `block[1]`, i.e. absolute field bit 50). The parent-38 section now correctly masks 30 fields (bits 39–49, 51–69).
- **Value serialization:** drop the `buf.write_float(sc.shield_block_crit_pct)` at the bit-50 position in the parent-38 write sequence.
- **Test:** update `active_player_stats_values_update_matches_cpp_common_runtime_masks` to assert the corrected mask bytes (`FF FB FF F0`) and walk past 30 parent-38 fields (not 31).
- **New regression test** `active_player_stats_values_update_omits_reserved_bit_50` asserts `block[1]` bit 18 is clear while bit 17 (ShieldBlock) and bit 19 (Mastery) are set — locking in the reserved-hole invariant.

## Grammar reference

In the 54261 client grammar, field 50 is never masked and never written: the mask-build loop has no `SetBit` for it and the value-write loop has no branch for it. TrinityCore's `UpdateFields.h` defines `ShieldBlockCritPercentage` for its own client target (where the field exists); the 54261 client tombstones the slot with no property, so the server must not emit it.

## Bit arithmetic

Field 50 → `block[1]` bit `50 % 32 = 18`; `0xFFFF_FFC0 & ~(1 << 18) = 0xFFFB_FFC0`; combined with the parent-0 bits 4,5 → `0xFFFB_FFF0`, big-endian `FF FB FF F0`.

## Testing

`cargo test -p wow-packet --lib active_player` — 15 tests pass (including the new regression).

### DIFF
--- a/crates/wow-packet/src/packets/update.rs
+++ b/crates/wow-packet/src/packets/update.rs
@@ -8157,9 +8157,14 @@ fn write_active_player_data_values_update(
         blocks[0] |= 1 << 0;
         blocks[1] |= (1 << 4) | (1 << 5);
 
-        // Parent 38 section: ALL 31 fields (bits 39-69)
-        // parent=38→b1:6, bits 39-63→b1:7-31, bits 64-69→b2:0-5
-        blocks[1] |= 0xFFFF_FFC0; // bits 6-31
+        // Parent 38 section: 30 fields (bits 39-49, 51-69). Field bit 50
+        // (ShieldBlockCritPercentage) is RESERVED in the 3.4.3.54261 client
+        // grammar: it has no property and is neither masked nor written (oracle
+        // hp_ObjectUpdateBuilder.cs:567,841 skip it). Emitting it shifts every
+        // following ActivePlayerData field by +4 bytes and desyncs the client's
+        // value walk, crashing the client shortly after world entry.
+        // parent=38→b1:6, bits 39-63→b1:7-31 EXCEPT bit 50→b1:18, bits 64-69→b2:0-5
+        blocks[1] |= 0xFFFB_FFC0; // bits 6-31 except bit 18 (field 50, reserved)
         blocks[2] |= 0x3F; // bits 0-5
 
         // Parent 269 section (block 8): SpellCritPercentage[7] + ModDamageDonePos[7]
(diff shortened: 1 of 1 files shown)


alseif0x:
Thanks a lot for digging into this and opening the PR.

I reviewed it against our canonical C++ source before merging. At the moment I cannot merge this as-is because the local TrinityCore/WotLK C++ source does define and write this field:

- `src/server/game/Entities/Object/Updates/UpdateFields.h:667-668` defines `ShieldBlock` and `ShieldBlockCritPercentage`.
- `src/server/game/Entities/Object/Updates/UpdateFields.cpp:2921-2922` writes both in `ActivePlayerData::WriteCreate`.
- `src/server/game/Entities/Object/Updates/UpdateFields.cpp:3672/3676` writes both in `ActivePlayerData::WriteUpdate` when changed.

So from the port discipline side, dropping bit 50 would currently be a deviation from C++ unless we can prove the 3.4.3.54261 client grammar differs from this C++ snapshot in this path. If you can attach a packet/client trace, a generated 54261 field layout, or a minimal repro showing C++ does not emit this field for the same VALUES update, we can revisit quickly.

Separately, your hotfix observation looks valid and very useful. Current `develop` already replies `Invalid(3)` instead of `RecordRemoved(2)` when Rust cannot serve a blob, and I added commit `297fc6f1` to avoid advertising Valid hotfixes that Rust cannot actually serve via SQL `hotfix_blob`. That should prevent the client from deleting local DB2 rows due to missing server-side blobs.

Keeping this PR open for now pending stronger evidence on the reserved field. Thanks again for the investigation.

alseif0x:
Follow-up after re-review: you are right that this is worth taking.

The nuance is that the C++ source does define `ShieldBlockCritPercentage`, but our current represented stats VALUES writer is not yet the fully generic C++ changes-mask writer. It emits a broad common stats section, and in that concrete Rust path bit 50 shifts the following ActivePlayerData values by four bytes for the 54261 client. Your regression test captures that failure mode cleanly.

I ran:

```bash
PROTOC=/home/cdmonio/.local/protoc/bin/protoc cargo test -p wow-packet active_player -- --nocapture
```

Result: 15/0 passing on the PR branch.

I'm going to squash-merge this. The follow-up on our side is to keep working toward the full 1:1 generic ActivePlayerData changes-mask writer, so this doesn't become a permanent broad-section shortcut. Thanks again, this was useful.

PR #5, opened 2026-06-24
fix(wow-data): serve only pre-decoded hotfix_blob records (fix client ERROR#132 on world entry)

jarlbrak:
## Problem

With a populated `hotfixes` DB, the real 3.4.3.54261 client crashes on world entry with **ERROR#132** (access violation, write, at `WowClassic.exe+0x17006e7` — the client's `WowDB2Lookup` table-index grow/rebuild).

Root cause: `HotfixBlobCache` advertises (`SMSG_AVAILABLE_HOTFIXES`) and serves (`SMSG_HOTFIX_CONNECT`) `Valid` hotfix records whose blob is sourced from `.db2` files loaded via `Wdc4Reader::record_bytes` — i.e. **raw on-disk WDC4 bit-packed bytes**, not the decoded field layout the client expects. (TrinityCore serves these from its in-memory store via `DB2StorageBase::WriteRecord`, ref `src/server/game/Handlers/HotfixHandler.cpp` `HandleHotfixRequest`.) The per-record `Size` is therefore wrong — e.g. `SpellMisc` 72 B raw vs 88 B decoded, `ChrCustomizationChoice` 12 B vs ~40 B — which desyncs the client's `HotfixConnect` content cursor. After the first record, every subsequent record is read at the wrong offset; one of those bogus reads lands on an array/record count, and the client rebuilds a DB2 table index with a negative count → crash.

This is **data-dependent**: it only triggers when `hotfix_data` is populated (e.g. from a TDB import). An empty `hotfix_data` declares nothing and never hits it.

## Fix

Serve only records backed by a `hotfix_blob` row. Those are pre-decoded wire bytes (produced by TC's hotfix population tooling) and are the correct format; `.db2`-file records loaded via `load_db2` are raw bit-packed and are not safe to serve as hotfixes.

- Track `hotfix_blob`-sourced records in a `hotfix_blob_keys` set, populated in `insert_blob` (used by `load_hotfix_blobs_from_db`; `load_db2` inserts directly into `blobs` and bypasses it).
- `load_hotfix_data_from_db` keeps a record only when it is `Valid` **and** backed by a `hotfix_blob` row (`is_servable_hotfix_record`). Raw `.db2`-only `Valid`, plus `RecordRemoved`/`Invalid`/`NotPublic`, are dropped — so `AvailableHotfixes` declare == serve, and the client uses its own correct local DB2 copy for the rest. This matches a real 54261 server, which serves only `Valid` records it can fully encode.

## Verification

- New positive/negative unit tests in `hotfix_cache.rs` (source tracking + servability gate).
- `cargo fmt --all -- --check`, `cargo test -p wow-data hotfix --lib`, `cargo check -p world-server` all pass; `cargo clippy` adds no new warnings on the changed file.
- **Real-client tested** (3.4.3.54261, populated `hotfixes` DB): world entry now reaches and holds in-world (~40-minute session, clean logout). Previously it ERROR#132'd within ~2 s of entering world. `HotfixConnect` went from 1927 records / 34 KB (raw, crashing) to 131 records / 7 KB (pre-decoded, stable).

## Note / follow-up

This is the correct **interim** behavior given there is no `WriteRecord`-equivalent WDC4 field decoder yet. Full parity (serving the complete decoded DB2 store like TC) would require implementing that decoder (Bit-packed / Pallet / Common / etc.), after which all DB2 records could be served. Until then, deferring to the client's local DB2 for non-`hotfix_blob` records is correct and crash-free.

### DIFF
--- a/crates/wow-data/src/hotfix_cache.rs
+++ b/crates/wow-data/src/hotfix_cache.rs
@@ -12,6 +12,7 @@
 
 use std::collections::BTreeMap;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::fs;
 use std::path::Path;
 
@@ -82,6 +83,12 @@ pub struct HotfixBlobCache {
     /// Inner key: record_id.
     /// Value: raw record bytes (inline strings, no copy-table dedup).
     blobs: HashMap<u32, HashMap<u32, Vec<u8>>>,
+    /// `(table_hash, record_id)` pairs whose blob came from the `hotfix_blob`
+    /// DB table (TC-population pre-decoded wire bytes) rather than a raw `.db2`
+    /// file (WDC4 bitpacked). Only these are safe to serve in
+    /// SMSG_HOTFIX_CONNECT — raw `.db2` bytes have the wrong size/shape and
+    /// desync the client's content cursor (ERROR#132).
+    hotfix_blob_keys: HashSet<(u32, i32)>,
     hotfix_data: BTreeMap<i32, HotfixPush>,
     optional_data: HashMap<String, HashMap<(u32, i32), Vec<HotfixOptionalData>>>,
     max_hotfix_id: i32,
@@ -122,6 +129,10 @@ impl HotfixBlobCache {
             .entry(table_hash)
             .or_default()
             .insert(record_id as u32, bytes);
+        // Mark this record as `hotfix_blob`-sourced (pre-decoded). `.db2` files
+        // are loaded via `load_db2` which inserts directly into `blobs` and
+        // bypasses this method, so they are NOT marked servable.
+        self.hotfix_blob_keys.insert((table_hash, record_id));
     }
 
     /// Load C++ `hotfix_blob` rows from the Hotfix database for one locale.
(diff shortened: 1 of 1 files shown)


alseif0x:
Thanks for the PR. We already covered this hotfix-blob issue on the current `develop` branch, but with a slightly different implementation to stay closer to TrinityCore/C++.

The important distinction is that C++ does not simply advertise/serve only records backed by `hotfix_blob`. For known DB2 stores, C++ keeps the `hotfix_data` record advertised and later tries `DB2StorageBase::WriteRecord`; `hotfix_blob` is only the fallback path for hashes without typed storage. If the requested valid hotfix cannot be serialized/served for the current locale, C++ downgrades the response status instead of sending malformed raw DB2 bytes.

Current `develop` now follows that shape:

- local raw DB2/WDC4 bytes are kept separate from SQL `hotfix_blob` payloads;
- SQL `hotfix_blob` rows are used only when they are actual pre-decoded hotfix payloads;
- raw DB2 bytes are no longer sent as `Valid` hotfix/DBReply payloads;
- known-store hotfixes are still advertised like C++;
- until Rust has a typed `DB2StorageBase::WriteRecord` equivalent for a store, the request path fails closed by downgrading the response instead of corrupting the client hotfix stream.

This is already included in `develop` together with the later LFG hotfix overlay work and the checkpoint commit `b5320412` (`checkpoint: align LFG hotfixes and random movement`). Because this PR is based on an older branch and now conflicts with `develop`, merging it as-is would regress the newer 1:1 behavior.

Current project state: we are continuing the C++ -> Rust port with a strict 0-gaps rule. The LFG hotfix overlay is implemented and `SMSG_LFG_PLAYER_INFO` has been byte-compared against C++; the active work now is fixing the remaining client crash on world entry by comparing Rust packet/runtime behavior against C++ and completing any missing dependencies immediately rather than leaving partial implementations.

alseif0x:
If you prefer, we can also repair the PR branch for you. Since `maintainerCanModify` is enabled, we should be able to rebase `fix/hotfix-serve-only-decoded-blobs` onto current `develop`, keep the current `develop` hotfix implementation, and force-push the updated branch so you can continue from a clean base.

Just confirm here if you want us to do that.

jarlbrak:
Closing — superseded by `297fc6f` ("Avoid advertising unserved hotfix blobs") on `develop`, which addresses the same SMSG_HOTFIX_CONNECT crash (raw WDC4/`.db2` record bytes are not the client wire format) more cleanly via a dedicated `hotfix_blobs` map. Glad we landed on the same root cause — thanks!

PR #6, opened 2026-06-26
fix(wow-data): only grant rewarded spells for skills the character has

jarlbrak:
## Problem

`SkillStore::starting_spells` decides which DBC-rewarded spells a character should auto-know. The gate

```rust
if !is_this_class_skill && !character_has_skill { continue; }
```

let any class-exclusive skill (`class_mask` with exactly one bit, matching the player's class) bypass the `character_has_skill` check. So a character was granted **every trainer rank of every single-class ability**, regardless of what is in its `character_skills`. A fresh level-1 character received ~352 spells where a real 54261 server sends ~52.

## Fix

Gate solely on `character_has_skill`:

```rust
if !character_has_skill { continue; }
```

This matches C++ `Player::LearnSkillRewardedSpells` (`src/server/game/Entities/Player/Player.cpp`), which is only invoked for skills the character holds — those are added first by `Player::LearnDefaultSkills` (race/class defaults, gated by `HasSkill` + level). `is_this_class_skill` is still used below to decide trainer-rank granting for skills the character *does* have, so their full rank set is unchanged. The `None` (no-filter) path is unchanged — `character_has_skill` is then `true` — preserving existing behavior and tests.

## Tests

- New hermetic `starting_spells_only_grants_class_skills_the_character_has`: a class-exclusive skill the character lacks is **not** granted; the same skill when known **is** granted; `None` still grants (backward compat).
- Existing `starting_spells` tests unchanged and pass. `cargo fmt --all -- --check`, `cargo test -p wow-data`, and `cargo clippy` are clean on the changed file.

### DIFF
--- a/crates/wow-data/src/skill.rs
+++ b/crates/wow-data/src/skill.rs
@@ -917,8 +917,18 @@ impl SkillStore {
                 .map(|ids| ids.contains(&skill_id))
                 .unwrap_or(true); // If no filter provided, allow all (backward compat)
 
-            if !is_this_class_skill && !character_has_skill {
-                continue; // Skip: profession/other-class skill the character doesn't have
+            // Only grant rewarded spells for skills the character actually has.
+            // The previous `!is_this_class_skill && !character_has_skill` gate let
+            // class-exclusive skills bypass `character_has_skill`, so a character
+            // with a populated `character_skills` set still received every trainer
+            // rank of every single-class ability — e.g. a fresh level-1 character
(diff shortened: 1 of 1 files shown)


jarlbrak:
Closing — superseded on `develop`: the `starting_spells` rework already removes the trainer-learned (`acquire_method=0`) DBC over-grant, and intentionally keeps the `is_this_class_skill` gate that this PR removed. Thanks!

PR #105, opened 2026-07-14
[08.4] [M0.7] Persist direct inventory swaps atomically

### LINKED ISSUE #104: [08.4] [M0.7] Persist direct inventory swaps atomically
Parent: #20

## Defect

D-C4: `handle_swap_inv_item` mutates the represented runtime first and then executes one `UPD_CHAR_INVENTORY_SLOT` per item independently while discarding both results. A failure after either update can leave the database partially swapped or divergent from memory. The explicit `CMSG_AUTO_EQUIP_ITEM_SLOT` direct-inventory path also calls the represented swap without persistence.

## C++ source of truth

- `src/server/game/Handlers/ItemHandler.cpp`: `HandleSwapInvItemOpcode`, `HandleAutoEquipItemSlotOpcode`, and the direct branch of `HandleSwapItem` validate the packet/position and call `Player::SwapItem`.
- `src/server/game/Entities/Player/Player.cpp`: `Player::SwapItem` performs the coherent in-memory move.
- `Player::_SaveInventory(CharacterDatabaseTransaction)` appends changed item/inventory positions to the character transaction rather than issuing independent best-effort slot updates.

## Rust targets

- `crates/wow-world/src/handlers/character.rs`
- Focused tests in the same module
- D-track documentation closeout

## Acceptance

- Empty-destination moves and occupied-slot swaps append all `character_inventory` slot changes to one `SqlTransaction`.
- The represented inventory, item-mod state, accessor/registry state, and client update packets change only after a successful DB commit.
- A missing/failed character DB commit leaves runtime state unchanged and reports an inventory failure.
- Every direct-inventory handler path that uses this represented swap receives the same persistence boundary, including explicit auto-equip-slot.
- Focused positive/negative tests cover the transaction plan and commit-failure rollback boundary.
- Validate formatting, focused `wow-world` tests, `world-server` check, local PR preflight/reviewer harness, and a complete bot relog persistence smoke before merge.

## Boundary

Full C++ container/bag validation, merge/fill behavior, and remaining `Player::SwapItem` gates are not expanded here; those remain tracked by #52.

Refs #20.

alseif0x:
## What changed

- persist direct-inventory swaps with one character transaction using TrinityCore-compatible `CHAR_REP_INVENTORY_ITEM` / `REPLACE INTO` semantics
- expose runtime slot, equipment-modifier, registry, loot-release, and player-value changes only after the database commit succeeds
- reject stale or malformed swaps whose source slot is empty, matching `Player::SwapItem`
- add an isolated bot workflow covering occupied-slot swap, logout, full re-authentication/relogin, inverse swap, and both persistence checks

## Why

The previous implementation issued independent position writes and could leave inventory state partially persisted when the second write failed. Occupied swaps also cannot be represented safely as two ordinary updates because `character_inventory.uk_location` requires each `(guid, bag, slot)` location to remain unique.

## Validation

- full local PR preflight: format, core checks/builds, focused library suites, and capture-diff
- `wow-world --lib`: 2,764 passed, 0 failed
- local Codex committed-patch review: CLEAN (0.91)
- release-build bot smoke: forward occupied swap persisted, relog passed, reverse swap persisted
- manual client QA: direct backpack swap remained in its final positions after relog

Closes #104

### DIFF
--- a/crates/wow-world/src/handlers/character.rs
+++ b/crates/wow-world/src/handlers/character.rs
@@ -110,6 +110,34 @@ const GOSSIP_OPTION_NPC_TRAINER_LIKE_CPP: u8 = 3;
 const GOSSIP_OPTION_TRAINER_TEXT_LIKE_CPP: &str = "I would like to train.";
 const ITEM_ENCHANTMENT_DB_FIELDS: usize = 3;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct DirectInventoryPositionUpdateLikeCpp {
+    slot: u8,
+    item_db_guid: u64,
+}
+
+fn plan_direct_inventory_swap_persistence_like_cpp(
+    src: u8,
+    dst: u8,
+    src_item: Option<&InventoryItem>,
+    dst_item: Option<&InventoryItem>,
+) -> Vec<DirectInventoryPositionUpdateLikeCpp> {
+    let mut updates = Vec::with_capacity(2);
+    if let Some(item) = src_item {
+        updates.push(DirectInventoryPositionUpdateLikeCpp {
+            slot: dst,
+            item_db_guid: item.db_guid,
+        });
+    }
+    if let Some(item) = dst_item {
+        updates.push(DirectInventoryPositionUpdateLikeCpp {
+            slot: src,
+            item_db_guid: item.db_guid,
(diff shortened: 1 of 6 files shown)


alseif0x:
@codex review

Issue #144, opened 2026-07-30
[ARCH.2.1a-pre2] Validate trainer load inputs before activation


alseif0x:
Parent: #133. Depends on #143 and #148. **Blocks #142** and must merge after #143 and before activating `TrainerBuySpell`.

## Why this prerequisite exists

The #143 C++ contrast replaced per-request trainer SQL reads with the already loaded process-wide `TrainerStoreLikeCpp`, so trainer list and the dormant buy handler now share one immutable catalog as C++ does. A final audit found that the Rust loader does not yet apply every relationship validation performed by `ObjectMgr::LoadTrainers` and `ObjectMgr::LoadCreatureTrainers`.

This is a pre-existing loader defect, not a reason to restore hot-path SQL. It becomes safety-critical before #142 makes `TrainerBuySpell` reachable: an invalid `trainer_spell` row can otherwise be listed, charged and persisted even though C++ would discard it at startup.

## C++ source anchors

- `src/server/game/Globals/ObjectMgr.cpp:9254-9304` (`ObjectMgr::LoadTrainers`):
  - rejects a missing `SpellId`;
  - rejects a nonzero missing `ReqSkillLine`;
  - rejects the row if any nonzero `ReqAbility1..3` spell is missing;
  - groups only validated spells and later reports rows whose `TrainerId` does not exist.
- `ObjectMgr.cpp:9329-9352`: ignores invalid/default locales and locales for missing trainers.
- `ObjectMgr.cpp:9364-9405` (`ObjectMgr::LoadCreatureTrainers`):
  - rejects a missing creature template;
  - rejects a missing trainer;
  - when either menu or option is nonzero, requires the exact gossip menu/order-index option;
  - publishes only validated `(CreatureID, MenuID, OptionID) -> TrainerID` mappings.
- `Entities/Creature/Trainer.cpp` / `Trainer.h`: runtime list/buy consumes the already validated immutable trainer catalog; it does not revalidate catalog foreign keys per request.

## Current Rust defect

`crates/wow-data/src/trainer.rs::TrainerStoreLikeCpp::from_rows_like_cpp` currently:

- accepts every `trainer_spell` row before checking the spell, skill-line or required-spell catalogs;
- validates only that a `creature_trainer` references an existing trainer;
- does not validate its creature template or selected gossip option.

The existing missing-trainer and locale checks are only a subset of the C++ loader contract.

## Goal

[...]

Issue #142, opened 2026-07-30
[ARCH.2.1a] Reconcile world dispatcher and registration drift


alseif0x:
Parent: #133. Depends on #135, #143, #144 and the ordered gameplay blockers #156, #163, #164, #157, #158, #159, #160, #161. Must merge after #161 and before #154/#134.

## Defect

The two sources required for an active world handler have drifted:

- `TrainerBuySpell` is present in `inventory::submit!` but absent from `WorldSession::dispatch_packet`; it passes the dispatch-table lookup and then falls into the wildcard trace branch without invoking `handle_trainer_buy_spell`.
- `PartyUninvite` and `MoveSetVehicleRecIdAck` have concrete dispatcher arms but no `PacketHandlerEntry`; the earlier dispatch-table lookup rejects them, so their arms are unreachable.

This was exposed while addressing the PR #141 review. It is a pre-existing gameplay defect, not part of the mechanical guardrail change.

## C++ anchors

`/home/server/woltk-trinity-legacy/src/server/game/Server/Protocol/Opcodes.cpp`:

- line 675: `CMSG_MOVE_SET_VEHICLE_REC_ID_ACK` — `STATUS_LOGGEDIN`, `PROCESS_THREADSAFE`, `HandleMoveSetVehicleRecAck`;
- line 714: `CMSG_PARTY_UNINVITE` — `STATUS_LOGGEDIN`, `PROCESS_THREADUNSAFE`, `HandlePartyUninviteOpcode`;
- line 978: `CMSG_TRAINER_BUY_SPELL` — `STATUS_LOGGEDIN`, `PROCESS_INPLACE`, `HandleTrainerBuySpellOpcode`.

## Blocking C++ contrast finding (2026-07-30)

Kickoff contrast found that C++ `PartyUninvite::Read` consumes `TargetGUID` through the packed `ObjectGuid` stream operator (`PartyPackets.cpp:114-123`, `ObjectGuid.cpp:779-785`), while Rust used a fixed 16-byte `read_guid()`. The missing registration kept this client-facing defect unreachable and its fixed-GUID test was self-consistent but not C++-shaped. Activating the opcode without correcting it would knowingly leave real client packets broken.

[...]

Issue #133, opened 2026-07-30
[ARCH] Modularize RustyCore: physical structure, ownership and extension seams


alseif0x:
## Current execution — 2026-09-07

**Global outcome unchanged:** a navigable modular Rust monolith with enforceable responsibility boundaries, canonical ownership, faithful base behavior and the approved bounded native/Wasm module acceptance. This epic defines the destination; **#584 is the operational coordinator and #585 is the selected next implementation delivery**. Do not copy the execution plan into every issue or reopen the entire architecture before each extraction.

PR #579 is merged into `3.4.3` at `59f5bced9b5cade0deb9150d82406f9113b4b61e`; #578 is closed for its canonical-owner foundation and represented lifecycle hardening. Action-specific live transfer/logout QA was explicitly deferred, not passed. Applicable finalization QA now belongs to #585; broader remaining acceptance stays in #584.

**Next: #585, represented session finalization outcomes and retirement supervision**, centered on `wow-world` with required consumers in `world-server` and narrow existing persistence/owner capabilities. Its closure does not complete C1, all C0–C4, full Login-side SaveToDB parity or this epic. No second operation family is selected yet. #585 contains the exact scope, exclusions, retirement list and acceptance; #584 contains the delivery rules and residual inventory.

Use one primary architecture implementation delivery at a time, coherent internal commits rather than helper issues, local bounded analysis, explicit separation of refactor/repair/product expansion, and full in-scope consumer migration. C0–C4 are acceptance dimensions applied to each responsibility, not five all-core prerequisite PRs. Physical splits need not wait for hecs, but cannot be presented as semantic ownership transfer. Introduced regressions and pre-existing defects invalidating the promised contract block that delivery; unrelated gaps retain named owners instead of automatically expanding it.

Preserve the chosen monolith/canonical-owner/private-selective-hecs/native-Wasm direction unless a demonstrated real-operation obstacle requires a bounded decision. No new generic framework, backend contest or analyzer campaign by default. The finite pre-migration conformance has passed within its recorded laboratory limits; production migration/SDK/durability acceptance remain separate.

[...]

PR #145, opened 2026-07-30
[ARCH.2.1a-pre] Model C++ interaction provenance

### LINKED ISSUE #143: [ARCH.2.1a-pre] Model C++ interaction provenance before activating TrainerBuySpell
Parent: #133. Depends on #135. **Blocks #142** and must merge before it.

## Why this prerequisite exists

The #142 kickoff contrast found that `TrainerBuySpell` is currently registered but unreachable because its dispatcher arm is missing. Activating that arm as-is would expose a security/behavior defect: Rust trusts the client-supplied `TrainerID` without proving that the player previously opened that exact trainer list.

This is not a speculative cleanup. C++ keeps one authoritative `PlayerMenu::InteractionData`; Rust currently represents only part of it as `gossip_source_guid` and separately mirrors the current banker. The missing `TrainerId` provenance makes the existing buy handler unsafe to activate.

## C++ source anchors

- `src/server/game/Entities/Creature/GossipDef.h:222-237`: `InteractionData { SourceGuid, TrainerId, PlayerChoiceId }`; `Reset()` clears all three.
- `GossipDef.cpp:234-238`: `ClearMenus()` clears menus only, not interaction provenance.
- `GossipDef.cpp:240-244`: `SendGossipMenu` resets provenance and stores `SourceGuid`.
- `GossipDef.cpp:307-313`: `SendCloseGossip` resets provenance before sending.
- `Handlers/NPCHandler.cpp:98-130`: trainer list validates the NPC; after resolving its ID, `SendTrainerList` removes feign death before trainer lookup, and only a valid trainer list replaces provenance with `{npc GUID, trainer ID}`.
- `NPCHandler.cpp:132-205`: trainer buy validates the NPC, removes feign death, then requires exact stored GUID and `uint32(packet.TrainerID)` before lookup or mutation; mismatch is silent and does not clear the binding.
- `NPCHandler.cpp:252-260`: gossip select requires the request GUID to equal stored `SourceGuid`.
- `Handlers/MiscHandler.cpp:1290-1297`: close interaction resets provenance only when `SourceGuid` matches.
- `Handlers/BankHandler.cpp:283-291`: showing the bank resets provenance and stores the banker GUID.
- `Entities/Player/Player.cpp:13739-13753`: trainer selection resolves the selected `(MenuID, OptionID)` trainer and falls back to the creature default when absent.

## Current Rust defects

- `WorldSession::gossip_source_guid` has no associated `TrainerId`.
- `handle_trainer_buy_spell` does not validate prior trainer-list provenance and skips NPC validation entirely when no canonical map manager is installed, despite the existing interaction helper supporting canonical and legacy maps.
- Trainer list/buy omit the C++ feign-death transition.
- `gossip_source_guid`, `gossip_options`, and `represented_current_banker_guid_like_cpp` mix or duplicate concepts that C++ keeps under one interaction authority plus a separate menu.
- gossip select falls back to the request GUID when no server-side interaction exists instead of requiring exact equality.
- binder emits `GossipComplete` without the C++ provenance reset.
- trainer gossip lookup lacks the target C++ specific→default trainer fallback; the current Rust unit test asserts the opposite behavior.

## Goal

Introduce one small, private, C++-shaped interaction-provenance owner in `WorldSession`, use it for the currently represented gossip/trainer/bank lifecycle, and make the dormant trainer-buy body safe and faithful before #142 makes it reachable.

## Scope

1. Replace `gossip_source_guid` and the duplicate current-banker GUID field with one private interaction-provenance value containing at least `SourceGuid` and `TrainerId`; keep `gossip_options` separate because C++ `Reset()` and `ClearMenus()` are different operations. Do not add a second trainer-specific mirror.
2. Encapsulate exact reset, generic-source bind, trainer bind, conditional close, and exact trainer-match operations. Compare `TrainerID` with C++'s `i32 as u32` bit-pattern semantics.
3. New gossip publication resets trainer provenance and stores only its source. Gossip select requires an active exact source GUID. Matching `CloseInteraction` resets provenance but does not pretend to be `ClearMenus`; a nonmatching close preserves it.
4. Add/use a server-side `SendCloseGossip`-equivalent that resets provenance before `GossipComplete`; route the existing binder close through it.
5. Route banker publication and bank authorization through the same source authority, removing the parallel banker mirror.
6. Trainer list must fail closed on NPC interaction through the existing canonical+legacy helper, remove feign death in C++ order after NPC/trainer-ID resolution, leave old provenance unchanged on failed trainer lookup, and atomically replace it immediately before publishing a successful list.
7. Restore the target C++ selected-trainer→default-trainer fallback so the provenance records the actual list sent.
8. Trainer buy must fail closed on NPC interaction in both runtimes, remove feign death after a valid NPC, require exact active `{SourceGuid, TrainerId}` before known-spell checks, DB access, packets, money or spell mutation, and retain the binding after success/failure for multiple purchases.
9. Update only focused tests and comments that encoded the old divergent behavior.

## Explicit boundaries

- The final loader-integrity audit is tracked by #144; it must land after this issue and before #142 activates the buy handler.
- Do not add the missing `TrainerBuySpell` dispatcher arm or touch handler registration/snapshot metadata; #142 owns activation.
- Do not redesign trainer price, prerequisites, `TeachSpell`, packet layouts, money transactions, persistence ordering, task ownership or scheduler behavior.
- Do not invent state resets for unrelated interaction UIs without an exact C++ producer anchor.
- `PlayerChoiceId` may be represented only if an existing Rust player-choice owner is migrated in the same one-owner model; otherwise document it as an unrepresented boundary instead of adding dead speculative state.

## Acceptance

- Unit tests prove reset, generic-source bind, trainer bind, exact GUID/ID matching (including signed-to-unsigned ID bit patterns), replacement and retention semantics.
- A successful resolved trainer-list publication stores the exact pair; a new gossip or matching close invalidates the trainer ID; failed lookup and nonmatching close preserve prior provenance.
- Buy tests prove no active binding, GUID mismatch and ID mismatch are silent before DB/money/spell work; exact binding reaches the pre-existing observable handler branch; valid NPC + mismatch removes feign death first, while invalid NPC does not.
- Gossip selection rejects absent/mismatched server provenance; bank uses the same source authority; `SendCloseGossip` resets it.
- Existing relevant gossip, bank, trainer, close-interaction, aura and packet tests pass.
- `cargo fmt`, `git diff --check`, full preflight, CI and the current-HEAD Codex reviewer verdict are green.

Refs #133, #142.

alseif0x:
Closes #143

Refs #133, #142, #144.

## What changed

- Replaced the separate gossip-source and banker mirrors with one private C++-shaped interaction owner containing `SourceGuid` and `TrainerId`, while keeping gossip menu entries separate like `PlayerMenu::ClearMenus`.
- Routed represented gossip publication/selection/close, direct services, bank authorization, trainer list and trainer buy through that single owner.
- Restored C++ ordering for source revalidation, fake-death removal, menu validation, trainer selected-to-default fallback, and conditional interaction close.
- Added fail-closed creature and GameObject gossip revalidation, including in-world, phase, icon-name and GameObject `OptionNpc::None` boundaries.
- Made trainer list/buy consume the process-wide immutable `TrainerStoreLikeCpp` already loaded at startup instead of issuing request-time world DB queries; corrected trainer wire states and localized greeting fallback.
- Added focused positive/negative regression coverage for provenance replacement/reset/retention, signed TrainerID semantics, gossip, bank, trainer and GO boundaries.

## C++ anchors

- `GossipDef.h:222-237`; `GossipDef.cpp:234-244,307-313`
- `NPCHandler.cpp:98-205,250-290`
- `BankHandler.cpp:60-75,283-291`
- `Player.cpp:1983-2007,13621-13699,13739-13753`
- `GameObject.cpp:2740-2756`
- `Trainer.cpp` / `Trainer.h`

## Explicit boundaries

- This PR does not add the dormant `TrainerBuySpell` dispatcher arm; #142 owns activation.
- The final audit found pre-existing missing `ObjectMgr::LoadTrainers` / `LoadCreatureTrainers` catalog validation. It is fully specified in #144, which now blocks #142 and must land next.
- Full trainer pricing/prerequisite/`TeachSpell` parity and a complete scripted GO gossip publisher remain outside #143.

## Validation

- `cargo +1.88.0 fmt --all -- --check`
- `git diff --check`
- `cargo +1.88.0 test -p wow-data trainer::tests --lib`: 6 passed
- `cargo +1.88.0 test -p wow-world --lib`: 3158 passed, 1 ignored, 0 failed
- `cargo +1.88.0 check -p world-server`: passed
- architecture guardrail self-test: 35 passed; repository handler contract: passed
- independent final C++ semantic audit: no actionable P1-P3 findings

`./tools/pr-preflight.sh full` passed architecture, format, core checks/builds/clippy and multiple focused suites, but the local Rust 1.88 toolchain later produced nondeterministic compiler SIGSEGV/ICEs (also reproduced on the unchanged base and in third-party crates on fresh targets). No Rust diagnostic or test failure was emitted; remote CI on the current HEAD is required as the clean-host authority.


alseif0x:
@codex review

Please review the current HEAD `a23b8f1e`, including the C++ ordering/provenance semantics and the explicit #144 loader-validation boundary.

PR #171, opened 2026-08-02
[ARCH.2.1a-pre4c] Make battle-pet ownership durable and account-atomic

### LINKED ISSUE #160: [ARCH.2.1a-pre4c] Make battle-pet ownership durable and account-atomic
Parent: #133. Blocking prerequisite for #142. Depends on #159.

## Defect

Rust's represented battle-pet journal is session-local. `battle_pet_add_pet_represented_like_cpp` uses a process-local counter, has no Login DB persistence, checks capacity outside the add operation, and cannot serialize two sessions of the same Battle.net account. The code says this limitation explicitly.

The target C++ also has a check/add race: `HasMaxPetCount` and `AddPet` are separate and `AddPet` does not revalidate the journal lock or capacity. We must not reproduce that race.

## C++ anchors and deliberate correction

- `BattlePetMgr.cpp:449+`: `AddPet`.
- `BattlePetMgr.cpp:567+`: per-species maximum.
- `BattlePetMgr.cpp:908+`: journal lock and mutation paths.
- Login DB `SEL/INS/UPD/DEL_BATTLE_PETS` and battle-pet slot statements.
- `BattlePetSpecies` flags for account-wide, owner-specific and legacy-account-unique species.

Rust must preserve journal/update semantics while moving the mutable collection to one shared owner per Battle.net account. Capacity, lock ownership, GUID allocation, durable insert and runtime publication become one result-bearing operation.

## Scope

- Introduce a narrow account-scoped battle-pet owner/facade shared by every world session for the same Battle.net account; do not expose its backing map.
- Load pets and slots from Login DB once coherently and allocate database-safe global battle-pet GUID counters.
- Make `try_add_pet` acquire/validate the account journal lease, recheck species flags and cap internally, compute stats, persist with an idempotency key, update the canonical account snapshot, then publish the complete inserted pet.
- Return structured failures for missing authority, lock ownership, invalid species, cap, duplicate request, DB failure and GUID collision.
- Route existing represented add/remove/update callers through the owner where practical without creating another mirror; define safe session attachment/detachment and shutdown behavior.
- Ensure no lock guard crosses `.await`; use a reservation/commit protocol or actor command with bounded completion.
- Do not charge trainer money or implement cross-DB trainer coordination in this PR.

## Acceptance

- Two sessions of one account cannot exceed the per-species cap or allocate duplicate pets; only the journal lease holder mutates.
- Account-wide and owner-specific species counts are exact, including legacy-account-unique max 1 and normal max 3.
- Insert/update/delete and slots survive restart/reload; failed persistence publishes nothing.
- Idempotent retry returns the same outcome without duplicate DB rows or packets.
- Initial `BattlePetUpdates` uses the complete canonical inserted object, including owner info.
- Existing battle-pet handler/spell tests are migrated to the facade and remain green.
- Focused multiclient/fault-injection tests, full preflight, CI and current-HEAD Codex verdict are green.

Refs #133, #142 and #159.

alseif0x:
Closes #160

## Summary
- move battle-pet ownership, slots, GUID allocation, receipts, and capacity enforcement into durable account-scoped Login DB state
- serialize account mutations across world processes and reload durable state during owner handoff
- make add/remove/update operations atomic with structured persistence failures
- reconcile uncage retries and guarded Character DB item destruction while preserving the C++ add, visual, DestroyItem ordering and gameplay side effects

## Scope boundary
Generic item-use opcode support remains outside this architecture issue; the temporary narrow handler work was reverted. The uncage effect path itself is durable and idempotent once invoked.

## Validation
- ./tools/pr-preflight.sh full origin/3.4.3
- committed capture-diff gate CLEAN
- local Codex review CLEAN


alseif0x:
@codex review

alseif0x:
Addressed every actionable review thread on current HEAD `c445780a`:

- the dedicated MySQL advisory-lock connection is monitored and process authority is revalidated;
- attach/detach/active-operation transitions are synchronized;
- every durable battle-pet mutation now validates a monotonically increasing per-account fence inside its transaction, preventing a former process from committing after handoff;
- successful persistence is revalidated before in-memory publication;
- uncage request keys are globally unique across accounts, with deterministic migration cleanup for historical cross-account duplicates;
- Battle.net account IDs use the unsigned domain and missing/zero links are rejected;
- invalid owner-shape rows no longer consume account-wide capacity;
- inactive account owners are identity-safely evicted.

Focused evidence: 21/21 battle-pet account tests pass, including a lease-loss/fence-handoff regression; formatting/diff checks and local Codex review are clean. The broad local preflight reached the focused-test stage but the host toolchain then produced transient rustc/ld SIGSEGVs and a nondeterministic handler-contract flake after that same contract had passed repeatedly, so remote clean-run CI is the definitive full gate.

@codex review

alseif0x:
Addressed all four findings in 5e234d8c:

- multiplexed account advisory locks through one process-level Login DB broker instead of one detached connection per online account; broker epoch invalidates all guards on connection loss and acquisition is cancellation-safe
- captured the durable fence atomically with the pending-add reservation, eliminating the orphaned reservation window
- scoped server-global lock names with a stable SHA-256-derived Login database identity
- verified/locked the pet row independently of MySQL changed-row counts, so no-op main-row updates can still replace declined names

Added focused regressions for DB lock-name isolation and declined-name-only replacement. Local Codex review is clean; `wow-world --lib` check passes.

@codex review

alseif0x:
@codex review

PR #207, opened 2026-08-18
[ARCH.PERF] Stop the architecture ratchet from dominating CI

### LINKED ISSUE #206: [ARCH.PERF] Make the architecture gate fast: release profile, one scan, incremental caches
`Check core crates` takes ~75 minutes, which makes every PR into `3.4.3` expensive
and blocks the modular-refactor track. This is tool cost, not build cost, and it is
almost entirely avoidable.

## Measured, not estimated

All figures are full workspace scans on the same machine, 337 source units.

| Configuration | Scan time |
|---|---:|
| Debug (what CI runs today) | ~2160 s |
| Release, unchanged code | ~600 s |
| Release + incremental caches | 490 s |
| Release + incremental caches + `debug-assertions` | **552 s** |

The job runs **two** full collections: `repository_surface_can_be_collected`
inside `cargo test`, and the ratchet command. (`handler-contract-check -- check`
is *not* a third — it only validates the handler contract TSV.)

## What was ruled out

Two obvious-looking optimizations were measured and rejected rather than shipped:

* **Caching the AST** — `syn::parse_file` runs 2022 times per scan (6 fixpoint
  iterations x 337 sources) which looks catastrophic, but it totals **10.1 s of
  ~600 s (1.7%)**. Parsing is not the cost.
* **Parallelizing the parse** — same reason.

## What is being changed

1. **Release profile for the scan steps.** 3.9x, measured. `debug-assertions` is
   turned back on in `[profile.release]` so the row-count cross-checks survive
   the profile change; it costs 11% and is worth it for a correctness gate.
2. **One full collection instead of two.** The ratchet step validates the same
   collected surface, so the duplicate repository test is skipped in CI.
3. **Incremental named-type caches.** `workspace_named_type_info_cache` and
   `package_named_type_info_cache` were rebuilt from scratch *per package, per
   fixpoint iteration* — 36 full rebuilds per scan, each deep-cloning the whole
   registry. They are pure functions of the registry and analyzing a package
   rewrites only that package's entries, so only the affected entries are
   refreshed. One redundant deep clone per iteration is also removed.
4. **Stop serializing the core build.** `-j1` was added to keep two large debug
   binaries from linking concurrently; it also serialized every dependency
   compile. Link one server at a time with bounded parallelism instead.

## Acceptance

The scan is deterministic, so correctness is binary and not a judgement call:
**the inventory must stay byte-identical** to the pre-optimization output
(12,127,560 bytes, reproduced across independent runs). Every change above was
verified against that oracle; any change that moves one byte is dropped.

## Known ceiling

Even with the gate at seconds, the job will stay ~15-20 min: the remainder is
compiling `wow-world`/`world-server` and two clippy passes. The goal is that the
ratchet stops being the dominant cost, not that CI becomes instant.

## Not included

Parallelizing the analysis across cores (the sources within a package are
independent) is deliberately left out. `collect_named_type_info` **assigns**
`type_generic_params` (last writer wins) while `VariableInfo::union` keeps the
*first* non-empty value, so a naive parallel merge would silently invert that
precedence. It is tractable — only one field is assigned, in two places — but it
deserves its own issue and its own adversarial tests.

alseif0x:
Closes #206.

`Check core crates` took ~75 minutes. The cost was the tool, not the build.

## Measured

Full workspace scans, 337 source units, same machine:

| Configuration | Scan |
|---|---:|
| Debug (what CI ran) | ~2160 s |
| Release, unchanged code | ~600 s |
| Release + incremental caches | 490 s |
| Release + incremental caches + `debug-assertions` | **552 s** |

End-to-end after the change:

* ratchet `check`: **575 s**, `PASS`
* tool test suite in release: **45.5 s**, 304 passed — it was **3380 s** in debug,
  because one test duplicated the full scan

Total scanning in the job: **~92 min to ~10 min**.

## Changes

1. **Release profile for the scan steps** (3.9x). `debug-assertions = true` is set
   in `[profile.release]` so the semantic row-count cross-checks survive the
   profile change — 11% slower, kept deliberately for a correctness gate.
2. **One collection instead of two.** The ratchet step validates the same surface
   `repository_surface_can_be_collected` collects, so the duplicate is skipped in CI.
3. **Incremental named-type caches.** `workspace_named_type_info_cache` and
   `package_named_type_info_cache` were rebuilt from scratch per package per
   fixpoint iteration — 36 full rebuilds per scan, each deep-cloning the whole
   registry. They are pure functions of the registry, and analyzing a package
   rewrites only that package's entries, so only affected entries are refreshed.
   The merged workspace view is recomputed only for consumers that declare the
   changed provider. One redundant deep clone per iteration is also removed.
4. **`-j1` removed.** It protected the concurrent link of two large debug binaries
   but serialized every dependency compile with it; the servers are now built one
   at a time with bounded parallelism.

## Correctness

The scan is deterministic, so this is binary rather than a judgement call: the
inventory must stay **byte-identical**. The pre-optimization output was reproduced
across two independent runs (12,127,560 bytes) and every configuration above was
compared against it. All identical.

The refreshed snapshot is not a behavioural change — it is the inventory the
analyzer merged in #203 already produced, which `3.4.3` never had installed:
23,404 rows, 13,048 production and 10,356 test-fixture. The 905 semantic groups
are **unchanged**, so the reviewed annotations already cover every new row. This
is what returns `Check core crates` to green on `3.4.3`.

## Measured and rejected

* **AST caching** — 2022 `syn::parse_file` calls per scan looks catastrophic, but
  totals **10.1 s of ~600 s (1.7%)**. Not worth doing.
* **Parallel parsing** — same reason.

## Deliberately out of scope

Parallelizing the analysis across cores. The sources within a package are
independent, so it is tractable, but `collect_named_type_info` **assigns**
`type_generic_params` (last writer wins) while `VariableInfo::union` keeps the
*first* non-empty value — a naive parallel merge would silently invert that
precedence, and today's workspace might not even expose it. That deserves its own
issue and adversarial tests rather than a quiet ride here.

## Known ceiling

The job will still take ~15-20 min: the rest is compiling `wow-world`/`world-server`
and two clippy passes. The goal was that the ratchet stops being the dominant cost.


alseif0x:
Fixed in `a2657945`. The self-test pins CI's exact commands precisely so the local gate cannot drift, and I changed the workflow without updating its mirror — that is what failed `Format`, before the optimized scan could even start.

Both halves now agree: the local `ci` profile runs the same release commands and skips the same duplicated collection, and the self-test expects the release forms plus the split per-server build. `./tools/pr-preflight.sh self-test` passes locally.

@codex review

alseif0x:
Both fixed in `c9b98a78`, and both were real gaps in the invariant rather than style points.

**Skip argument** — the expectation ended at the manifest path, so dropping `-- --skip repository_surface_can_be_collected` from either side would have left `self-test` green while restoring the duplicate full-workspace scan this PR exists to remove. Since both sides wrap that argument onto a continuation line, it gets its own named invariant rather than a longer single-line string.

**Split build** — pinned only `bnet-server`. The split exists so two large debug binaries never link concurrently, so pinning half of it would let the other half be dropped or recombined unnoticed. Both halves are now pinned in the dry-run output and the workflow text.

Verified adversarially rather than assumed: removing the skip fails with `GitHub workflow single-collection skip appeared 0 time(s)`, and removing the world-server build fails with `GitHub workflow world-server linked build appeared 0 time(s)`. Restored, self-test green.

@codex review

PR #209, opened 2026-08-18
[ARCH.PERF.1] Drop gate work that enforces nothing

### LINKED ISSUE #208: [ARCH.PERF.1] Cut the gate's dead weight: redundant check, decorative clippy, unused scan
Follow-up to #206. A multi-lens audit of the architecture gate found work that costs CI
minutes while enforcing nothing. This issue is the zero-risk subset: **no guarantee is lost
by any item here.**

## 1. The `cargo check` of six crates is redundant

`.github/workflows/rust-ci.yml` checks `wow-data wow-database wow-network wow-world
bnet-server world-server`, then builds `bnet-server` and `world-server`. Verified from
`crates/world-server/Cargo.toml`: world-server depends directly on all four libraries, and
bnet-server on `wow-database`. The builds already type-check everything that line covers.

Keep the separate `wow-test-bot` check — it is outside the workspace.

**~2-3 min.**

## 2. Both clippy steps cannot fail

Verified, three independent ways:

* `[workspace.lints.clippy]` (`Cargo.toml:185-193`) sets everything to `warn`, but **no crate
  declares `lints.workspace = true`**, so the table is inherited by nothing.
* No `#![deny(...)]` or `#![forbid(...)]` in any of the five crates involved.
* No `-D warnings` anywhere in the workflow, and `:165` passes `--cap-lints warn`, which
  degrades an explicit `-D` back to a warning.

They cost 1.1 min and are guaranteed to exit 0 whatever they find.

Rather than delete, **repurpose**: a `clippy.toml` deny-list on `disallowed_types` /
`disallowed_methods` covering sqlx query/pool/transaction items and the concrete
`wow-database` types, applied to crates that should hold no persistence. Same 1.1 min, and it
becomes the first rule in this family that actually *prohibits* something rather than
recording that it happened.

Note the honest limit: the deny-list cannot point at `wow-world` (2,227 production
operational rows) or `wow-data` (1,551) without a large allow-list, so it protects the crates
that are already clean. That is still strictly more than today's zero.

## 3. `print-baseline` runs an 11-minute scan and throws it away

`print_repository_baseline` (`session_ownership.rs:2097`) calls `collect_repository_baseline`,
which performs the full workspace persistence collection — to emit a struct whose persistence
field is `#[serde(skip)]` (`:214-215`). This is a straight bug, not a trade-off.

Give the session-syntax path a collector that skips the persistence collection. **0 CI min,
~11 min off every local baseline regeneration** — and it makes it cheap to un-skip
`repository_surface_can_be_collected`, currently the only test that cross-checks the collector
against an independent `syn::parse_file`.

## 4. The hotspot ratchet globs the whole workspace to watch 8 files

`check_architecture.py:2250-2254` walks every `crates/*/src/**/*.rs` through a pure-Python Rust
lexer, while `validate_hotspot_non_growth` (`:2290-2320`) reads only the 8 audited paths in
`runtime-ownership-ledger.json`. Restrict the walk to those paths; keep the presence check that
catches renames.

Also: the validator block runs twice — `pr-preflight.sh:872` (`self-test`, Format job) and
`rust-ci.yml:92` (`check`).

**~0.25 min in `Check core crates`, ~0.25 in `Format`.**

## 5. Dead fields in the frozen identity

`classification` is a pure function of the package name (`session_ownership.rs:1778-1784`) and
**no rule reads it** — `compare_persistence_access_baseline` runs first and returns on `?`.
`cfg` is perfectly redundant with `source_class`: only two combinations exist across all
23,404 rows. Dropping both leaves exactly 23,404 distinct keys.

Artifact shrink, no CI time, no behaviour change.

## Done when

`Check core crates` is ~20 min from ~22.5, the clippy steps can fail on a real violation,
`print-baseline` no longer scans, and the snapshot carries no field that nothing reads.
The inventory must stay byte-identical apart from the two removed fields.

alseif0x:
Closes #208. Follow-up to #206.

Three cuts to `Check core crates`, **none of which costs a guarantee**.

## 1. The `cargo check` of six crates duplicated the builds beside it

Verified in `crates/world-server/Cargo.toml`: world-server depends directly on `wow-data`,
`wow-database`, `wow-network` and `wow-world`; bnet-server on `wow-database`. The two builds
already type-check everything that line covered. `wow-test-bot` keeps its own check — it is
outside the workspace and nothing else builds it. **~2-3 min.**

## 2. Neither clippy pass could fail

Verified three independent ways:

* `[workspace.lints.clippy]` (`Cargo.toml:185-193`) exists, but **no crate declares
  `lints.workspace = true`**, so nothing inherits it.
* No `#![deny(...)]` / `#![forbid(...)]` in any of the five crates.
* No `-D warnings` anywhere, and the world pass passes `--cap-lints warn`, which degrades an
  explicit `-D` back to a warning.

They cost every PR 1.1 min to print advice that gated nothing. Moved to the scheduled job so
the signal stays visible and free. Making clippy *enforcing* means addressing wow-world's lint
debt first, and deserves its own change.

## 3. `print-baseline` paid ~10 min for a value it discarded

`print_repository_baseline` ran the full workspace persistence collection to render an envelope
whose persistence field is `#[serde(skip)]` (`session_ownership.rs:214-215`). The collector now
takes it as a parameter. **~590 s to 6.3 s.**

Verified the result is only *stored*, never derived from: `persistence_accesses` appears twice
in `collect_units` (parameter, then passed to `finish`), and `finish` places it in the struct
literal at `:1651`. The rendered output is unchanged.

This also makes it cheap to un-skip `repository_surface_can_be_collected` later — currently the
only test cross-checking the collector against an independent `syn::parse_file`.

## Rejected after measuring

**A clippy deny-list on sqlx and the concrete `wow-database` types.** This was the plan, and it
does not work. `wow-loot`, `wow-entities`, `wow-map` and `wow-network` cannot *name* those types
because they cannot declare the dependency: `dependency-policy.json` puts them in the protected
`domain-runtime` category and allows `wow-loot` exactly `rand`. The deny-list would have
enforced nothing that the Cargo-edge check does not already enforce in milliseconds — enforcement
theatre on top of a gate whose problem is precisely that it records rather than prohibits.

**Restricting the hotspot glob** to the 8 audited paths. Measured: the whole Python guard is
14.2 s, so this saves ~10 s and costs the ranking report the refactor programme uses.

**The "duplicate" validator runs** (`pr-preflight.sh:872` self-test vs `rust-ci.yml:92` check)
are different modes in parallel jobs; they cost no wall clock.

## Validation

* Tool suite in release: **304 passed, 0 failed**
* `cargo fmt --all --check`, `git diff --check`: clean
* `./tools/pr-preflight.sh self-test`: passes, and the local profile is mirrored — the new
  invariant `CI profile must not spend PR time on clippy passes that cannot fail` was verified
  adversarially by reintroducing a clippy call and watching it fail.

Expected: `Check core crates` ~22.5 min to ~19-20 min. The remaining dominant cost is the
11.3-min exact scan, which #206's audit addresses in later steps.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -149,7 +149,11 @@ jobs:
 
       - name: Cargo check
         run: |
-          cargo +1.88.0 check --locked -p wow-data -p wow-database -p wow-network -p wow-world -p bnet-server -p world-server
+          # The two builds below already type-check these crates: world-server
+          # depends directly on wow-data, wow-database, wow-network and
+          # wow-world, and bnet-server on wow-database. Checking them first
+          # compiled the same crates twice. wow-test-bot stays: it is outside
+          # the workspace and nothing else builds it.
           cargo +1.88.0 check --locked --manifest-path tools/wow-test-bot/Cargo.toml
           # `-j1` was introduced to stop two large debug binaries from linking
           # concurrently, but it also serialized every dependency compile in the
@@ -158,11 +162,6 @@ jobs:
           cargo +1.88.0 build --locked -j4 -p bnet-server
           cargo +1.88.0 build --locked -j4 -p world-server
 
-      - name: Clippy loot authority boundary
-        run: cargo +1.88.0 clippy --locked --no-deps --message-format short -p wow-loot -p wow-entities -p wow-map -p wow-network --lib
-
-      - name: Clippy world target with inherited debt capped
-        run: cargo +1.88.0 clippy --locked --no-deps --message-format short -p wow-world --lib -- --cap-lints warn
 
   tests:
     name: Focused library tests
@@ -239,3 +238,13 @@ jobs:
         run: |
           cargo +stable check --locked -p bnet-server -p world-server
           cargo +stable build --locked -p bnet-server -p world-server
+
+      # Neither pass can fail: no crate sets `lints.workspace = true`, none
+      # denies a lint in source, there is no `-D warnings`, and the world
+      # target caps lints to warnings outright. They were costing every PR
+      # 1.1 minutes to print advice nobody was gated on. Keep the signal
+      # where it is free, and make it enforcing in its own change.
+      - name: Clippy advisory sweep
+        run: |
+          cargo +1.88.0 clippy --locked --no-deps --message-format short -p wow-loot -p wow-entities -p wow-map -p wow-network --lib
+          cargo +1.88.0 clippy --locked --no-deps --message-format short -p wow-world --lib -- --cap-lints warn
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -205,6 +205,49 @@ jobs:
           cargo +1.88.0 run --locked -p capture-diff -- verify-required loot-single-item-claim
           cargo +1.88.0 run --locked -p capture-diff -- verify-required creature-spell-casting
 
+  clippy:
+    name: Clippy advisory sweep
+    # Neither pass can fail a build: no crate sets `lints.workspace = true`, so
+    # the workspace lint table is inherited by nothing; no crate denies a lint
+    # in source; there is no `-D warnings`; and the world pass caps lints to
+    # warnings outright. That cost every PR 1.1 minutes to print advice which
+    # gated nothing, so the signal lives here instead — visible and free.
+    # Making it enforcing means clearing wow-world's lint debt first.
+    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
+    env:
+      CARGO_INCREMENTAL: "0"
+    steps:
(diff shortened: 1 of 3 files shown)


alseif0x:
Fixed in `445ba662` — you were right, and the failure mode was worse than a red job: being schedule-only, `toolchain '1.88.0' is not installed` would have gone unnoticed until someone went looking for clippy output that had never been produced.

Rather than switch the sweep to `+stable`, I gave it its own job on the project's pinned 1.88 with the clippy component. Folding an advisory lint sweep into a job whose purpose is testing a *different* toolchain conflated two things; the stable job should keep testing stable.

The preflight self-test caught a second-order effect on its own: the new job's `CARGO_INCREMENTAL: "0"` made that setting appear three times where the invariant pinned two. Expectation updated to three (check, tests, clippy).

`./tools/pr-preflight.sh self-test` passes, `cargo fmt --all --check` and `git diff --check` clean.

@codex review

PR #211, opened 2026-08-19
[ARCH.PERF.2] Ratchet the session surface on PRs, the exact inventory on merge

### LINKED ISSUE #210: [ARCH.PERF.2] Take the exact persistence scan off the PR path
Follow-up to #206 / #208.

After #209 the `Check core crates` job is 15m48s, and the exact persistence scan is **11.3 of
those minutes — 72% of the job**. Nothing else on the PR path comes close.

## The change

Run the exact persistence comparison on **push to `3.4.3` and on the weekly cron**, not on pull
requests. Keep everything else exactly where it is.

PRs keep: `Format`, `Focused library tests`, the handler-contract check, the Cargo-edge
dependency policy, capture-diff, the Codex review — and the session-ownership syntax ratchet,
which #209 already made separable from the persistence collection
(`collect_repository_baseline_with_persistence`) and costs ~6 s.

Expected: the job drops to **~4.5 min**, and since the four required checks run in parallel a
PR settles in about **5 minutes** rather than 16.

## What this costs, stated plainly

A new direct-persistence access is detected **at merge instead of at PR time**. It is not
detected *less* — every merge to `3.4.3` still runs the full comparison, and so does the cron.
What is lost is immediacy, measured in hours.

That trade is defensible here because the mechanism is a change detector on source text, and
it is at its weakest on a codebase in motion: 55% of the last 1,692 commits touch
`crates/wow-world/src/session.rs`, so most of what it reports during active development is
regeneration churn rather than a leak.

## Why not simply delete it

Because it is one day old and has not had a fair trial. Moving it keeps every bit of coverage
and can be reverted with a one-line change to the job's `if:` condition.

## Also considered and dropped

Restricting the collection to production sources. Verified byte-exact (13,048 rows identical to
the previous production subset; the 105 operational fixture rows reproducible with zero misses)
and it halves the scan — 552 s to 269 s. But it saves nothing once the scan is off the PR path,
and it would force a snapshot migration for a nightly job. The measurement worth keeping from
that experiment: **the emitting pass is ~91% of the per-class cost**, not the fixpoint, so
future work on this analyzer should target the emitting walk.

## Done when

A PR settles in ~5 min, `3.4.3` still fails on a new direct-persistence row, and the session
ownership ratchet still runs on every PR.

alseif0x:
Closes #210. Follow-up to #206 / #208.

After #209 the job is 15m48s and the exact persistence scan is **11.3 of those minutes — 72%**.

## What changes

Pull requests run `check --syntax-only`: the same session ownership ratchet without the
workspace persistence collection. **6.4 s instead of 11.3 min**, with every session number
still enforced — 727 production + 11 fixture `WorldSession` fields, 20 impl owners, 3339
associated items, 243 `SessionResources` fields, 685 registry rows.

The exact inventory runs on push to `3.4.3` and on the weekly cron.

Expected: the job drops to **~4.5 min**; since the four required checks run in parallel, a PR
settles in about **5 minutes** rather than 16.

## What it costs

A new direct-persistence access is caught **at merge instead of at review time**. Not caught
less — every push to `3.4.3` and every cron run does the full comparison. Reverting is a
one-line change to the step's `if:`.

The trade is defensible because this is a change detector on source text, which is weakest on a
codebase in motion: 55% of the last 1,692 commits touch `session.rs`, so most of what it reports
mid-development is regeneration churn rather than a leak.

The local preflight still runs **both** — catching an inventory move before pushing is the
difference between fixing it and reddening the branch.

## Invariants

Two new self-test expectations, both verified adversarially:

* removing the `if: github.event_name != 'pull_request'` fails with
  `post-merge-only exact persistence scan appeared 0 time(s)`
* dropping `--syntax-only` fails with `session-ownership syntax ratchet appeared 0 time(s)`

## Implemented, verified, then dropped

Restricting the collection to production sources. It works and is exact: 13,048 production rows
**identical** to the previous production subset, and all **105** operational fixture rows
reproduced with zero misses and zero extras. It halves the scan, 552 s to 269 s.

It is not here because it saves nothing once the scan is off the PR path, and it would force a
snapshot migration for a job that now runs nightly. The measurement worth keeping: **the
emitting pass is ~91% of the per-class cost**, not the fixpoint — future work on this analyzer
should target that walk, not the convergence loop.

## Validation

* Tool suite in release: 304 passed, 0 failed
* `check --syntax-only`: PASS in 6.36 s
* `cargo fmt --all --check`, `git diff --check`, `pr-preflight.sh self-test`: clean

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -99,12 +99,26 @@ jobs:
             -- --skip repository_surface_can_be_collected
           cargo +1.88.0 run --release --locked --manifest-path tools/architecture/handler-contract-check/Cargo.toml -- check
 
+      # Pull requests ratchet the session surface only — seconds. The exact
+      # persistence inventory is an 11-minute scan that was 72% of this job,
+      # and it is a change detector on source text: on a branch where 55% of
+      # commits touch session.rs, most of what it reports mid-development is
+      # regeneration churn. It still runs below on every push to 3.4.3 and on
+      # the cron, so a leak is caught at merge rather than at review time —
+      # deferred, not dropped.
+      - name: Ratchet the session ownership surface
+        if: github.event_name == 'pull_request'
+        run: |
+          cargo +1.88.0 run --release --locked --manifest-path tools/architecture/handler-contract-check/Cargo.toml \
+            --bin session-ownership-check -- check --syntax-only
+
       # The combined check compares the session surface before the persistence
       # snapshot, so a session-only mismatch must not launch the long scan
       # below. Publish the baseline only when the persistence comparison is
       # what failed.
       - name: Check persistence and session ownership ratchets
         id: ratchets
+        if: github.event_name != 'pull_request'
         run: |
           set +e
           report="$(cargo +1.88.0 run --release --locked --manifest-path tools/architecture/handler-contract-check/Cargo.toml --bin session-ownership-check -- check 2>&1)"
--- a/tools/architecture/handler-contract-check/src/bin/session-ownership-check.rs
+++ b/tools/architecture/handler-contract-check/src/bin/session-ownership-check.rs
@@ -8,6 +8,7 @@ use std::process::ExitCode;
 fn usage() -> ExitCode {
     eprintln!(
         "usage: session-ownership-check check [--policy PATH]\n       \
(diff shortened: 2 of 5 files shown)


alseif0x:
@codex review

Issue #186, opened 2026-08-11
[ARCH.DB.1] Inventory and ratchet concrete persistence leakage


alseif0x:
Parent: #169 and #133. Depends on #181. Followed by the persistence-order recorder and capability ports.

## Defect

Concrete SQL, `sqlx` types, public pool access, dialect-specific locks and transaction construction are spread through application/gameplay code. The current dependency policy lists broad exceptions but does not provide a call-site inventory with semantic owners and retirement conditions.

## Goal

Create a reproducible, production-vs-test-aware persistence boundary inventory and a non-growth guard before any port or adapter migration.

## Scope

- Inventory direct `sqlx` imports/types, `Database::pool()` consumers, raw/direct queries, interpolated SQL, explicit transactions, statement builders, generated-ID reads, advisory locks and multi-database workflows.
- Classify each site by logical database, capability owner, transaction/connection affinity, current order, error/unknown-commit behavior and target child issue.
- Separate production, test/fixture and generated code counts.
- Add executable ratchets that reject new unowned concrete persistence access and stale exceptions.
- Synchronize the architecture issue ledger/dependency policy with #169 and its children.
- Do not alter SQL, transactions, public behavior, schemas or backend selection in this issue.

## Acceptance

- The inventory is generated/reproducible from HEAD and every production exception names an open removal/decision issue.
- New unowned direct SQL/pool/sqlx access fails an adversarial fixture.
- Existing MariaDB behavior and tests are unchanged.
- Architecture self-tests/check, full preflight, CI and current-HEAD Codex verdict are green.

Refs #133, #169 and #181.

PR #212, opened 2026-08-19
[ARCH.DB.2] Record persistence plans as ordered semantic traces

### LINKED ISSUE #187: [ARCH.DB.2] Freeze Player lifecycle persistence order before its port
Parent: #169 and #133. Depends on completed #186. This focused contract is required by #200, not by physical splits or unrelated domain refactors.

## Problem

Player login/save/logout behavior depends on statement order, transaction boundaries, connection affinity, commit classification, fences and publication. Moving that workflow behind a port without an executable contract could preserve final rows while breaking retry/crash behavior.

The prior scope attempted to freeze Player, inventory, loot, group, quest and battle-pet flows with one universal recorder. That produced a multi-PR tooling project and is retired.

## Goal

Freeze only the current C++-anchored Player lifecycle persistence contract needed for #200.

## Scope

- Record stable semantic steps and typed parameter classes for load, periodic/manual save and logout.
- Record transaction begin/commit/rollback, logical DB/connection affinity, retry/fence, unknown-commit classification and publication after commit.
- Cover normal success, failure before commit, unknown outcome after commit and reconnect/logout boundary.
- Exercise real production entry points where feasible; use the existing recorder only where it provides value.
- Redact secrets and values not needed by the contract.
- Annotate Rust/C++ discrepancies without correcting them in the tracing PR.
- Do not build a process-wide generalized recorder, SQL tokenizer or cross-domain golden framework.
- Loot, Group and later capabilities carry focused contracts in their own port issues.

## Acceptance

- Deterministic Player lifecycle traces fail on reordered commit/publication or changed connection affinity.
- Positive, rollback and unknown-commit behavior is covered.
- No lock is held across external I/O in the traced path.
- The fixture is small enough to understand beside #200.
- #218 is superseded; transaction scoping/tokenization work is not required unless this focused contract proves it necessary.

Validation: tools/local-harness.sh final origin/3.4.3 plus focused MariaDB/lifecycle tests. Full persistence audit is optional unless explicitly requested.

alseif0x:
Refs #187. **Deliberately not `Closes`** — see *What is not here*.

Rows are not the persistence contract. Crash and retry behaviour is decided by statement order,
transaction boundaries, which logical database and connection carried a step, how a failed
`COMMIT` was classified, and what was published after it. Extracting ports can preserve every
final row and break all of that invisibly.

## What is here

`PersistenceTrace` — an ordered record with a JSON golden form, covering transaction begin,
statement, raw statement, advisory lock, commit, rollback, deadlock retry, fence and publication.

**Identity is semantic, never SQL text.** Statements are named by their statement-enum variant —
the analogue of C++'s `CharacterDatabaseStatements`. The exact inventory from #186 keys on source
text, so a rename or reformat moves it; a golden that behaved the same way would be useless during
the very refactor it exists to protect. Identity comes from `Debug`, which all four enums already
derive, so 1002 variants needed no boilerplate.

**Parameters are redacted by construction.** Numbers keep their value; text and blobs keep only
length and an FNV-1a digest. These plans bind account names, session keys and password verifiers,
so the trace must prove a value changed without storing it. Tested both ways: a secret never
appears in the rendered trace, and two different secrets still compare unequal.

**Nothing was wired by hand.** Seventy-five call sites build transactions. Threading a recorder
through them would have left the forgotten ones silently untraced — the worst failure for a
golden, since absent events read as absent persistence. `SqlTransaction::new` picks up an ambient
recorder instead. Zero call sites changed.

**Production pays one relaxed atomic load.** Capture happens only inside a `RecordingSession`.
A test asserts that with no recording installed, nothing is recorded at all.

Commit attempts are classified into three outcomes that are deliberately not interchangeable: a
definite rollback may be replayed, an unknown outcome may not. Collapsing `Unknown` into either
neighbour is the silent data-loss bug this contract exists to expose, and two tests pin it.

## Two defects found while building it

**Only transactions were instrumented.** Running the recorder against a real flow produced an
empty trace. Single statements go through `query`/`execute`, and a read that gates a write is not
incidental. The tell was that `ConnectionAffinity::Pooled` existed in the vocabulary and was never
emitted. Both paths now record before touching the pool, so tests cover them against a
deliberately unreachable database.

**The contract violated itself.** The local Codex review caught it: `CharStatements::GENERATED_CPP`
is a data-carrying variant holding its query, so its derived identity rendered as
`GENERATED_CPP { sql: "SELECT money FROM ..." }` — identified by SQL, the exact coupling this work
was introduced to avoid. The variant now carries its C++ identifier, and three assertions pin the
fix: the identity contains no SQL, survives a formatting-only change, and keeps two generated
statements distinguishable. The flaw was in the central claim, not at the edges.

## What is not here

The issue also asks for versioned goldens of six flows (player save/logout, inventory+money, loot,
group, quest, battle-pet) and fault injection around commit. Neither is in this PR.

The reason is measured, not assumed. I built a probe over the vendor-purchase flow — the one with
an existing atomicity test — and the trace came back **empty**: with the unreachable pool that test
uses, that flow never reaches SQL at all; its "definite rollback" happens in an in-memory tracker.
Choosing flows that genuinely exercise persistence needs another pass, and claiming coverage here
would be the bulk close AGENTS.md prohibits.

The recorder those goldens need is complete and tested. The follow-up is picking the flows and
freezing them.

## Validation

* `wow-database --lib`: 185 passed, 0 failed
* `wow-world --lib battle_pet`: 153 passed (needs `RUST_MIN_STACK=268435456`, as CI sets)
* `./tools/pr-preflight.sh full origin/3.4.3`: architecture, handler contract, session ownership
  and exact persistence ratchets clean; capture-diff CLEAN on both required flows. The local Codex
  step timed out at its own limit (exit 124) — a local tooling budget, not a finding.
* Inventory re-frozen twice as the code changed: **23,555 rows, 911 groups**. Six new production
  workflows needed reviewed annotations, and the policy generator refused to derive anything until
  they had them — the one rule in this family that makes a boundary a commitment.


alseif0x:
Nine findings, all correct. Two fixed here; the other seven are tracked in #213 with the reasoning quoted, because each needs a design decision rather than a patch.

**Fixed in `61fe5eaf` + `f0f40be7`:**

*Pre-COMMIT rollbacks.* You were right twice over. The first pass fixed one of three paths — `try_commit_inner` abandons the transaction before `tx.commit()` when it cannot acquire a connection, cannot execute a statement, or fails an affected-row check, and the last two used `?`, so they recorded nothing at all. A trace would have shown a transaction that began and never ended. All three now emit `Rollback`, and `Commit` is reserved for an attempt that actually reached `tx.commit()`. Only the first has a test — the other two need a reachable database that then fails — so they are covered by one shared helper on every early return until the MariaDB goldens exist.

*Float width.* `F32(0.0)` and `F64(0.0)` collapsed into one trace despite sqlx sending different MySQL type metadata. The width is now part of the record.

**Tracked in #213**, and I want to flag the P1 rather than bury it: `PreparedStatement::new(X.sql())` is the repository's common pattern and those statements carry no identity, so `record_statement` **silently omits them** — including `UPD_CHAR_MONEY` at `session.rs:29246`. An incomplete trace that looks complete is worse than none. It is not fixed here because it needs a decision about how those builders keep their identity, and because this PR delivers no goldens that could yet be misled by it.

Also in #213: `HotfixStatements::GENERATED_BASE` has the *identical* defect I fixed for `CharStatements::GENERATED_CPP` and I did not check it — same class, missed instance.

Inventory re-frozen four times as the code changed: 23,568 rows, 910 groups. Full preflight clean.

@codex review

alseif0x:
Sixteen findings across four rounds, all correct. The eight actionable ones are now closed, including both P1s; the rest are tracked in #213.

**Both P1s.** Manually built statements keep their identity — 85 call sites migrated to `PreparedStatement::for_statement(X)`, so a transaction whose first statement was manual no longer fails to open in the trace. And the three money workflows that open `pool().begin()` directly are recorded: boundary, statements and termination.

**The rest:** pre-COMMIT rollbacks at all three early returns; integer and float bind widths; pooled statements recorded after the call so a failure is distinguishable from a success; `MixedLogicalDatabases` as its own event; the advisory-lock lifetime, both halves; and generated hotfix statements identified by their table rather than their SQL.

On your last finding specifically — you were right, and it was the third time today I fixed one instance of a defect and assumed the rest. `reconcile_stored_item_money_commit_like_cpp` recorded a transaction that opened and never ended, which is precisely the defect I had just fixed for `SqlTransaction`. I verified the repair this time by counting each hook per function instead of trusting the edit:

```
attempt_group_loot_money_transaction_like_cpp    begin=1 statements=2 termination=2
attempt_stored_item_money_transaction_like_cpp   begin=1 statements=2 termination=3
reconcile_stored_item_money_commit_like_cpp      begin=1 statements=2 termination=1
```

**Still open, and said plainly in the module docs** so nobody freezes a golden on it: nine indirect statement builders, raw pooled SQL with no attributable database, and the honest limit that an unsalted digest of a low-entropy value is not proof against a dictionary — a trace is safe to read, not necessarily safe to publish.

Inventory re-frozen: 23,437 rows, 910 groups. Two hotspot ceilings raised deliberately, since #133 forbids moving code before it is encapsulated and this is the observation that has to exist first. Full preflight clean, including the local review.

@codex review

alseif0x:
All three P1s from the fourth round are closed.

**Durable mutations were invisible.** The stored-item transaction recorded its two `SELECT ... FOR UPDATE` reads and neither of its writes, so removing or reordering the credit or the source consumption would not have moved the trace. `UPD_CHAR_MONEY` and `DEL_ITEMCONTAINER_MONEY` are recorded now.

**Termination is structural, not remembered.** You were right that per-site hooks would keep missing returns — SQLx rolls an unfinished transaction back on drop, so every `?` ends it. All three money workflows now open an `ExplicitTransactionTrace`, an RAII guard that records the rollback if it drops unresolved, so a return added later is covered without anyone remembering to cover it. Verified by counting hooks per function rather than trusting the edit:

```
attempt_group_loot_money_transaction_like_cpp    guard=1 statements=2 termination=2
attempt_stored_item_money_transaction_like_cpp   guard=1 statements=4 termination=3
reconcile_stored_item_money_commit_like_cpp      guard=1 statements=2 termination=1
```

**The false rollback was mine.** A failed pooled `execute` emitted `Rollback`, but there is no transaction on that path and a transport error may have applied the write — the trace asserted a revert that may never have happened. That is a false record, which is worse than a missing one, and it was a defect I introduced fixing your previous round. It now records the ambiguity instead.

Also derived the group payout's update identity from `CharStatements::UPD_CHAR_MONEY` instead of hard-coding a name, so routing the same operation through `Database::prepare` cannot move a golden just because two paths spelled one statement differently.

**Four P2s remain open in #213**, with your evidence quoted: successful pooled affected-row counts, advisory-lock loss outside orderly shutdown, hotfix identities collapsing two queries on one table, and missing `DeadlockRetry` events for explicit transactions.

Inventory re-frozen: 23,434 rows, 910 groups. Full preflight clean on the first attempt.

One environmental note, since it affected how this was verified: this machine produced five single-bit memory corruptions during the session — one persisted to disk in `session.rs`, the rest transient in reads, plus repeated rustc segfaults. Every edit here was audited line by line against `git diff`, and the snapshot was checked against git rather than regenerated when a parse failed.

@codex review

alseif0x:
Final round from my side. Fixed the last finding and one thing it exposed.

**Definite versus ambiguous pooled failures.** You were right that my previous fix overshot: making every pooled failure ambiguous denied a definite fact, because acquiring a connection fails before MySQL receives anything. `PoolTimedOut`, `PoolClosed` and `Configuration` are definite non-executions; a protocol or I/O error after the send stays ambiguous. Both directions have tests now — asserting a revert that never happened and asserting ambiguity where none exists are the same defect with opposite signs, and I shipped each once.

**A stack budget, not a code defect.** The inventory scan then segfaulted fifteen consecutive times and I concluded the machine could no longer run it. That was wrong. It needed `RUST_MIN_STACK=1073741824`; my classification code deepened the analyzer's recursive AST walk past the 256 MiB CI budget. Raised in the workflow and the preflight, because since #211 that scan runs post-merge — a stack overflow there would have reddened `3.4.3` for every later PR with no signal on the PR that caused it.

Worth naming how it was found: I had a ready explanation (this session produced six genuine single-bit memory corruptions, two persisted to disk) and stopped looking. Capturing stderr instead of discarding it found the real cause in one command. Same mistake as the retry loop that swallowed two `E0308` errors earlier — suppressing output to keep a loop tidy hides the reason.

**Where this stands:** 21 of 26 findings across six rounds are closed, including all four P1s. Inventory 23,463 rows / 911 groups, pair verified, full preflight clean.

The five remaining are design changes, not omissions — per-task recording state, transaction correlation ids, normalized raw-SQL shape, salted redaction, nine indirect builders — and they are named in the module documentation so nobody freezes a golden on them.

I am not opening a seventh round. The maintainer's priority is the modular refactor, and this PR has already cost more than the contract it delivers is worth. If the remaining items block the verdict, they should be weighed against landing a working recorder rather than fixed here.

Refs #187, #213.

alseif0x:
@codex review

Reviewing `c6cb3368`. Since the last round: definite-vs-ambiguous pooled failures are separated with tests in both directions, and CI's stack budget is raised to 1 GiB because the inventory analyzer segfaults below it — that one would have reddened `3.4.3` post-merge with no signal on the PR that caused it.

The five design-level findings still open are consciously parked in #213, closed as not-planned with the reasoning recorded, and named in `persistence_trace.rs`'s module documentation so nobody freezes a golden on a gap they cannot see: process-wide rather than per-task recording state, no transaction correlation ids, raw SQL digested from exact bytes, unsalted parameter digests, and nine indirect statement builders.

That is a deliberate scope decision by the maintainer, not an oversight.

alseif0x:
@codex review

Reviewing `7e32308d`. Since your last pass, two of your three findings are fixed:

The explicit hooks no longer build anything when tracing is off — they take a closure, and the guard is inert without a recorder. Your point was exact: the group payout was allocating a String, a Vec and taking a mutex per recipient on a production money path while tracing was disabled, contradicting the single-atomic-load cost the module claims. A test now asserts the closure does not run in the production configuration, which is the only way that claim stays true.

`deny_unknown_fields` is on the trace and event schemas, so dropping a field can no longer let an old golden parse and compare equal.

The cancellation finding is deliberately not fixed and is recorded as such: a pooled call cancelled mid-await never reaches the post-await recorder. Fixing it needs two events per statement — attempt and completion — which changes the trace shape. It is a visible gap rather than a false record, and the module documentation names it.

One correction to my own earlier claim in this PR: I raised CI's stack budget and explained it as a stack-overflow fix. That was wrong — the crash reproduces with an unlimited stack and with the changes reverted, so its cause was the authoring machine's memory fault and the one success at 1 GiB was luck. The budget stays as harmless headroom; the comment now says what is actually known rather than the tidy explanation I reached for.

alseif0x:
@codex review

Reviewing `7e32308d`. Two of your three findings from the previous pass are fixed:

The explicit hooks no longer construct anything when tracing is off — they take a closure and the guard is inert without a recorder. Your point was exact: the group payout was allocating a String, a Vec and taking a mutex per recipient on a production money path while tracing was disabled, contradicting the single-atomic-load cost the module documents. A test asserts the closure does not run in the production configuration.

`deny_unknown_fields` is on the trace and event schemas, so dropping a field can no longer let an old golden parse and compare equal.

The cancellation finding is deliberately left open and named in the module docs: a pooled call cancelled mid-await never reaches the post-await recorder. Fixing it needs two events per statement, which changes the trace shape; it is a visible gap, not a false record.

This commit also corrects a claim I made earlier in this PR. I raised CI's stack budget and explained it as a stack-overflow fix; that was wrong. The crash reproduces with an unlimited stack and with the changes reverted, so its cause was the authoring machine's memory fault and the single success at 1 GiB was luck. I used one positive result to overturn fifteen negatives because it fit a tidy explanation. The budget stays as harmless headroom and the comment now says what is actually known.

alseif0x:
@codex review

alseif0x:
## Triage of the 19 findings

The rebase onto the fixed gate triggered a fresh review of the whole PR rather
than of one commit, and a large share of it re-raises work already in the branch.
I verified each against the code instead of fixing blind, because fixing what is
already there would have churned the diff without changing behaviour.

### Fixed (5)

| Finding | Commit |
|---|---|
| **P1** Raise the analyzer's main-thread stack | `0d5895df` |
| **P1** Instrument post-commit publication events | `4d828f90` |
| Raw SQL dropped when the transaction has no attributed database | `0a556855` |
| Defer transaction events until the batch is executed | `b807b6c2` |
| Avoid deterministic hashes of secret parameters | `61ec70c8` |

Two of these deserve more than a table row.

**The stack finding was right and it corrected a wrong diagnosis of mine.**
`RUST_MIN_STACK` sizes stacks for threads Rust spawns, and this analyzer spawned
none — it called the scan from `main`. So the 1 GiB budget in CI, and the commit
that justified it, reached nothing. `print-persistence-baseline` had been
segfaulting on every attempt on my checkout, including under
`RUST_MIN_STACK=1073741824` and `ulimit -s unlimited`, and including on a
pristine `3.4.3` worktree. I had concluded the machine's memory was failing. On a
`std::thread::Builder` with an explicit stack it completes: exit 0, 12 MB, 10m18s.

**The abandoned-batch finding changed my mind mid-review.** I had judged
append-time recording intentional — the PR records *plans*, and C++ separates
`trans->Append` from `CommitTransaction` the same way. The vendor-currency
turn-in defeats that: it appends its statements and returns when it cannot take
the money lock, so `pool.begin()` is never reached and the trace described writes
that never happened. `BatchAbandoned` now marks it, deliberately not `Rollback` —
nothing was sent, so nothing was undone.

### Already in the branch (11)

Verified present, with the line that shows it:

- **Float bind width** — `Float { bits, width_bits }`; the doc comment already
  gives the `F32(0.0)` / `F64(0.0)` example the finding proposes.
- **Integer bind widths** — `Int { value, width_bits }`, `Uint { value, width_bits }`.
- **Pooled affected-row counts** — `result.as_ref().ok().map(|r| r.rows_affected())`.
- **Mixed logical databases** — `PersistenceEvent::MixedLogicalDatabases { opened, appended }`, with a test.
- **Advisory-lock loss outside shutdown** — `record_advisory_lock(&lock_name, false)` in *both* monitor loss branches.
- **Advisory-lock lifecycle** — the event is constructed in production at four sites, acquisition included.
- **Reject unknown fields when parsing goldens** — `deny_unknown_fields` on both the event enum and the trace struct.
- **Generated hotfix identities** — `hotfix_base_table_like_cpp` / `hotfix_projection_shape_like_cpp`.
- **P1 Preserve identity for manually built statements** — the two sites named do not do this: `group.rs:1023` already uses `for_statement`, and `session.rs:29246` is an unrelated function. All eight occurrences of `PreparedStatement::new(<enum>.sql())` are inside `#[cfg(test)]`. The residue underneath was real and is `0a556855`.
- **P1 Trace explicit SQLx transactions** — `attempt_group_loot_money_transaction_like_cpp` and `attempt_stored_item_money_transaction_like_cpp` both carry `ExplicitTransactionTrace` hooks.

### Deferred, with reasons (3)

These are real and none is a small change. Per this issue's own scope —
*"characterize current Rust behavior and annotate any known C++ discrepancy;
correcting parity is a separate issue/PR"* — they belong to the follow-up that
builds the flow harness, not to the recorder that precedes it.

- **Correlate events with their transaction.** Concurrent same-database
  transactions interleave with no identifier, so `begin-A, begin-B, commit-B,
  commit-A` cannot be grouped. Correct, and it needs a per-transaction identity
  threaded through every event — a schema change worth making once the harness
  exists to exercise it.
- **Workflow-local recording state.** The ambient recorder is process-wide by
  deliberate design: 75 call sites build transactions, and threading a recorder
  through all of them would dwarf the contract it serves, leaving the ones nobody
  updated silently untraced. Scoping it to a task is the right end state and is
  the same change as the item above.
- **Hash a normalized raw-SQL shape.** Correct that byte-hashing recreates
  text-coupling. A naive normalization would corrupt string literals, so this
  needs a small tokenizer and its own tests rather than a regex.

`cargo test -p wow-database --lib`: 199 passed, 0 failed. `cargo check -p
wow-world`: clean. Exact inventory re-frozen with all of the above in.

alseif0x:
@codex review

alseif0x:
All threads addressed. Which commit each landed in, or where it went instead.

| Finding | Resolution |
|---|---|
| **P1** Raise the analyzer's main-thread stack | `0d5895df` |
| **P1** Instrument post-commit publication events | `4d828f90` |
| **P1** Replace the stale C# statement authority | `1fa6c898` |
| Raw SQL dropped when no database is attributed | `0a556855` |
| Defer transaction events until the batch is executed | `b807b6c2` |
| Preserve in-flight batch outcomes on cancellation | `2cb7d1fe` |
| Avoid deterministic hashes of secret parameters | `61ec70c8` |
| Record bound parameters for raw prepared statements | `2f6bf517` |
| Preserve terminal outcomes for raw-only transactions | `26f0bc8d` + `fa48cf37` |
| Bypass the ambient mutex when tracing is disabled | `9aefd0ed` |
| Restore the outer recorder after nested sessions | `9aefd0ed` |
| Do not report cancelled COMMITs as rollbacks | `dff25bc1` |
| Align the frozen order with the post-await recorder | `1fa6c898` |
| Reopen the traced boundary for each deadlock retry | `1fa6c898` |
| Skip explicit trace construction when recording is disabled | stale — both hooks take a closure and test `recording_enabled()` first |
| Float / integer bind widths, pooled row counts, mixed databases, advisory-lock lifecycle and loss, `deny_unknown_fields`, generated hotfix identities, manual statement identity, explicit SQLx transactions | already in the branch; verified against the code, listed with their evidence in the earlier triage |
| Correlate events with their transaction · Use workflow-local recording state · Hash a normalized raw-SQL shape | deferred to **#218**, with acceptance criteria |

Three deserve more than a row.

**The ambient mutex** was the worst of them and it was mine: every
`Database::query` and `Database::execute` locked a process-wide mutex to discover
there was nothing to record. A global serialization point on the normal
production path, in the one mode this facility is supposed to cost nothing.

**Cancelled COMMITs** appeared twice — once in `SqlTransaction`, then again in
`ExplicitTransactionTrace`, the sibling I had not checked after fixing the first.
Both now separate "nothing was sent" from "COMMIT went out and the answer never
came", because a golden that calls the second a rollback lets a retry duplicate
the write.

**Raw-only transactions** were the deepest. Making the statement visible left the
transaction invisible: no begin, no commit, no rollback, for flows like the
bank-slot purchase and the tutorial save. Fixing it needed the attribution to
become a property of the *type* — `Database<S>` holds no `S` to ask — and the
audit grammar then rejected the associated `const` I reached for and named the
surface it wanted. It is `fn database()`.

One correction to the record: the snapshot in `941374d4` was generated with the
analyzer binary built from #215's worktree, which carries a splice this branch
does not, and froze eighteen rows that do not belong here. `9a8a22db` re-freezes
from this branch's own build at **23,486 rows**. What caught it was reading the
names of the rows that disappeared on the next scan rather than the total.

`cargo test -p wow-database --lib`: 207 passed, 0 failed. `cargo check -p
wow-world -p world-server`: clean. Exact ratchet: PASS.

@codex review

alseif0x:
Nine findings, six fixed and three deferred.

| Finding | Resolution |
|---|---|
| Record only the packet publications that were enqueued | `8711bece` |
| Record advisory-lock release when RELEASE_LOCK errors | `8711bece` |
| Replay statement events inside each retry boundary | `8711bece` |
| Use ambient tracing in `SqlTransaction::default` | `8711bece` |
| Represent pooled non-execution without a rollback | `8711bece` |
| Refresh the documented inventory totals | `8711bece` |
| Attribute pooled execution to the receiving database | **#218** |
| Make `with_trace` preserve semantic statement identities | **#218** |
| Distinguish generated hotfix projections of equal width | **#218** |

**Five of the six were regressions from earlier fixes in this PR.** The clearest
is the retry boundary: last round I gave each deadlock retry its own
`TransactionBegin` and stopped there, so the trace read as a begin followed
straight by a commit while the retry re-sent every statement — a worse account of
the attempt than the missing boundary it replaced. It replays them now.

`SqlTransaction::default` is the one I would not have found. Derived, it left
`trace: None` unlike `new()`, and the legacy-password migration replaces its
batch with `std::mem::take` every ten thousand accounts: first batch traced,
every later one silently not. A trace that looks complete and covers a fraction
of the work is the failure mode this whole contract exists to prevent.

The three deferred are pre-existing rather than introduced here, and each needs
the flow harness that does not exist yet. #218 carries them with acceptance
criteria.

`cargo test -p wow-database --lib`: 208 passed, 0 failed. `cargo check -p
wow-world`: clean. Both guards clean, inventory re-frozen at 23,505 rows with the
two new workflows annotated.

@codex review

alseif0x:
Three findings, all three on the previous round's own fixes. Fixed in `47cf4d80`
— **not resolving the threads this time**: four of the last five findings landed
on lines I had marked resolved, so the verdict is the check, not my say-so.

**Publication.** Gating on `journal_enqueued && learned_enqueued` traded one
wrong trace for another: when exactly one packet goes out, the client has
observed a publication and the trace recorded none — partial delivery now looked
like a closed channel instead of looking like success. The client observes them
separately, so they are recorded separately.

**BEGIN failure.** `pool.begin()` failing called `record_rollback`, the same
terminal event as an opened transaction that executed and failed. Nothing was
sent in the first case, and that is the distinction a retry turns on. It records
`BatchAbandoned`, and the test — which drives an unreachable pool — now asserts
both directions.

**Totals.** 23,505 rows, 12,978 production and 10,527 test-fixture, multiplicity
25,743. The previous update was written before the final re-freeze, so it was
stale on arrival; this one is taken from the snapshot as pushed.

`cargo test -p wow-database --lib`: 208 passed, 0 failed. `cargo check -p
wow-world`: clean. Architecture guard clean, and the exact inventory is
unchanged at 23,505 — these three fixes move no rows.

@codex review

alseif0x:
Fixed in `3afa8c65` and `056f9abe`.

**Wrong adapter.** `SqlTransaction` erases the statement family, so a Character batch compiles when handed to `Database<LoginStatements>` — the SQL runs on Login while the trace says Character. `attribute_to_like_cpp` returned early on an already-attributed batch and hid exactly the contradiction that recording connection affinity is for. It records `MixedLogicalDatabases`, the same event an in-batch mismatch already produced.

**BEGIN failure in the explicit workflows.** The guard is constructed only after `pool().begin()` succeeds, so a connection that could not be acquired left the trace empty — a definite non-execution reading like the workflow never being reached. Both money paths record `BatchAbandoned` before returning.

**Deferred to #218:** emitting `TransactionBegin` only once BEGIN has succeeded. Statements are recorded as they are appended, so the begin is emitted then too; moving it means buffering every event until execution starts, which is the same change as the workflow-scoped recorder already tracked there rather than a local fix.

208 → 209 tests, `cargo check -p wow-world` clean, both guards clean, inventory re-frozen at 23,510.

@codex review

alseif0x:
Fixed in `dbb43232`.

The purchase enqueues the player-money values update **before** the Character commit, and the trace carried only the later journal and learned-spell packets. Moving that packet across the commit, or after the pet packets, produced an identical trace — and that ordering is the crash window the event exists to pin. The client sees the charge before the pet, so it is recorded where it happens.

**Delaying the begin until the transaction opens** stays deferred to #218, now written up there with acceptance criteria. Statements are recorded as they are appended, so the boundary is emitted then too; moving it means buffering every event until execution starts, which is the same change as the workflow-scoped recorder already tracked there. Doing it piecemeal is what produced the current shape — the raw-statement buffer added for attribution already holds *some* events, and a second partial rule would leave two orderings in one recorder.

`cargo check -p wow-world` clean, architecture guard clean, exact inventory unchanged at 23,510.

@codex review

alseif0x:
Fixed in `84c16a98` — same mistake as the first publication hook, one function further along.

`send_player_values_update_from_entity_bridge` returns `()` and returns early when there is no player GUID or values snapshot, so recording unconditionally claimed the client saw a charge that was never enqueued. The bridge now reports whether it sent and the hook is gated on that; its other four callers ignore the value, as they already did.

`cargo check -p wow-world` clean, architecture guard clean after raising session.rs by the six production lines the return and its comment add.

@codex review

alseif0x:
Fixed in `0e20abc8`.

`send_packet` logged and discarded a closed-channel error, so the bridge returning `true` after calling it meant only that it got that far — the money publication was gated on a value that could not fail, which is the same wrong trace the gate was added to prevent, one level down. It returns whether the channel accepted the packet now. Its several hundred other callers ignore the value as before; two match arms in `spell_acquisition::application` needed braces because they tail-returned it.

**Deferred to #218:** isolating the ambient recorder from tests running in parallel. `capture_flag_test_lock` serializes recording tests against each other, but an ordinary test that builds a transaction still writes into the active recording — that is the workflow-scoped recorder already tracked there, not a local fix.

Guard, exact ratchet and formatter clean; session.rs raised by ten production lines and one signature re-frozen out of 3,339.

@codex review

alseif0x:
Of the four findings: the reversed comment is fixed in `b0322d4d`. The other
three are on **#218**.

The comment did claim the opposite of what the code does — the charge is awaited
and matched before the packet is enqueued, so the publication is after the
Character commit, not before. In a file whose whole subject is ordering, a
comment with the order backwards is worse than no comment.

The refund publication, the cancelled pooled write and the missing
`DeadlockRetry` on explicitly traced transactions are all the same shape as
what this PR already added, and each one added so far has produced the next:
gating a publication revealed the enqueue could fail, and fixing that revealed
`send_packet` swallowed the failure. They need one pass that makes
attempt-and-outcome uniform across the pooled, batched and explicit paths, not
a seventh hook bolted onto this PR. That is #218 items 5–7, with acceptance
criteria.

What this PR freezes is sound as far as it goes: statement order, transaction
boundaries, connection affinity, commit/unknown classification, fences and the
publications on the paths it covers. The gaps are recorded, not forgotten.

@codex review

Issue #150, opened 2026-07-30
[ARCH.2.4a] Introduce an opaque PlayerRegistry facade in place


alseif0x:
Parent: #133. Historical prerequisites #181/#185 are closed; this slice is ready. Followed by #192–#196 and relocation #138.

## Defect

`wow-network::PlayerRegistry` is a public `DashMap<ObjectGuid, PlayerBroadcastInfo>`. Callers obtain entry guards and invoke storage operations directly, bypassing incarnation, recipient and capability invariants. Moving that alias would only relocate the defect.

## Goal

Introduce a named, opaque connected-session directory at its current location and migrate only the lifecycle identity operations needed to establish the seam. Encapsulation precedes consumer migration and physical relocation.

## Scope

- Replace the public type alias with a newtype whose storage is private to its module.
- Add narrow operations for register/replace, generation-aware lookup, unregister and control-channel addressing.
- Migrate the production lifecycle callers covered by those operations.
- Return owned handles/snapshots, never mutable `DashMap` guards, iterators, entry APIs or a generic mutation closure.
- Use the #181 inventory to assign every remaining consumer exactly once to #192 runtime/fanout, #193 combat/loot, #194 quest/spell/movement, #195 social/group/chat or #196 compatibility-fixture closure.
- Preserve the 80-field `PlayerBroadcastInfo` temporarily as explicit mirror debt; reject field growth.
- Do not move the module, redesign mailbox commands, introduce an event bus or migrate Player/Map state in this PR.

## Invariants

- Same registration replacement, stale-session rejection, lookup, unregister, channel identity and shutdown behavior.
- Exactly one directory owner and one current session incarnation per entry.
- No new mirror, unbounded queue, lock across await or service-locator API.

## Acceptance

- Lifecycle production callers use named operations and storage is inaccessible through the public type.
- Remaining compatibility access is owner-private, counted and mapped to #192–#196.
- Tests cover replacement incarnation, stale generation, absent registration and unregister races.
- Direct-access and 80-field ratchets cannot increase.
- Validation follows docs/operations/local-first-development.md: tools/local-harness.sh final plus the focused evidence required above. Remote checks follow #222; full preflight and model review are opt-in.

Refs #133, #138, #181, #185 and #192–#196.

PR #215, opened 2026-08-19
[MOD.1] Extract root test modules from the eight audited hotspots

### LINKED ISSUE #214: [ARCH.2.0] Extract the test modules that make the hotspots monstrous
Parent: #133. This lands before the encapsulation work in P1 (#136, #150, #151) because it is
mechanical, carries no behavioural risk, and removes the majority of the volume that makes those
files hard to work in.

## The measurement

The hotspot files are more than half tests:

| File | Total | Production | Tests |
|---|---:|---:|---:|
| `crates/wow-world/src/session.rs` | 166,815 | 71,927 | **94,888 (57%)** |
| `crates/wow-map/src/map.rs` | 33,660 | 15,245 | **18,415 (55%)** |
| `crates/wow-world/src/handlers/loot.rs` | 29,950 | 13,716 | **16,234 (54%)** |
| `crates/wow-world/src/handlers/misc.rs` | 18,788 | 7,315 | **11,473 (61%)** |
| `crates/wow-world/src/handlers/quest.rs` | 18,472 | 8,255 | **10,217 (55%)** |

`session.rs` holds a single `#[cfg(test)] mod tests` of 94,772 lines. Extracting it leaves the
file at **72,046** — a 57% reduction with not one production line touched.

## Why this is not "moving before encapsulating"

#133 forbids moving code before its invariants are encapsulated, because a move that only
relocates public aliases leaves the invariants scattered. Moving *tests* relocates no invariant:
the production module boundary, its visibility and its owners are untouched. The repository
already uses this pattern — `#[path = "character_vendor_atomicity_tests.rs"]` in
`handlers/character.rs`.

## A measurement bug this exposes

`check_architecture.py` partitions lines by `#[cfg(test)]` extents *inside* a file. A test module
extracted to its own file carries that attribute on the `mod` declaration in the parent, so the
scan finds none and counts all 94k lines as **production**.

Uncorrected, this inverts the metric the refactor is steered by: moving 94k test lines out of a
hotspot would report the hotspot unchanged and a new 94k production file appearing. The
classifier needs to recognise a whole-file test module via an inner `#![cfg(test)]`.

## Scope

* Extract the root `mod tests` from each file above into a sibling `*_tests.rs` included with
  `#[cfg(test)] #[path = ...] mod tests;`.
* Teach the hotspot classifier about whole-file test modules, with a self-test fixture.
* Lower the affected hotspot ceilings in `runtime-ownership-ledger.json` to the new production
  counts — these are upper bounds, and this is the rare change that genuinely reduces them.
* Refresh the hotspot table in `docs/architecture/ownership-and-boundaries.md`.

## Done when

Every extracted file compiles, the full suite passes with an identical test count, the hotspot
report attributes the extracted lines to tests, and `session.rs` is under 73,000 lines.

## Also worth adjusting in #133

The epic states that LOC is a diagnostic rather than a success criterion, but its hotspot table
sums production and test lines into one Total, which is what makes these files read as
"monstrous". Consider reporting production lines per ownership unit instead, so that the real
work — splitting ownership — is not measured together with test placement.

alseif0x:
Closes #214

First slice of the modularization work. Each of the eight audited hotspots keeps
its production code exactly where it is and moves its root `mod tests` to a
sibling `#[cfg(test)] #[path = "..._tests.rs"] mod tests;`.

| Hotspot | Before | After | Sibling |
|---|---:|---:|---:|
| `session.rs` | 166,769 | 72,000 | 94,042 |
| `map.rs` | 33,660 | 15,250 | 18,264 |
| `character.rs` | 30,853 | 20,238 | 10,480 |
| `loot.rs` | 29,853 | 13,663 | 16,033 |
| `main.rs` | 28,175 | 15,383 | 12,516 |
| `misc.rs` | 18,788 | 7,318 | 11,315 |
| `quest.rs` | 18,472 | 8,260 | 10,163 |
| `player.rs` | 18,172 | 9,268 | 8,884 |

This is test *placement*, deliberately not ownership progress: no production
code moves, no test moves between crates, and the retirement conditions in the
ledger are untouched. The hotspots stop being unopenable in an editor, which is
the prerequisite for the actual submodule split.

### Why the hotspot classifier changed too

The first commit teaches the classifier that a file reachable only under
`cfg(test)` counts as test lines. Without it, a pure test move would have been
scored as a ~110k-line production regression and the ratchet would have
rejected it. The gate stays closed in the other direction as well: the sibling
only counts as test lines when the *entire* file is cfg(test)-gated, so
production code cannot be parked in a sibling to shrink a hotspot.

### Verification

`cargo check` does not compile tests, so it cannot see damage on either side of
the split. Three independent checks instead:

- **Production is byte-identical.** Diffing pre- vs post-extraction production
  lines gives zero removals and zero modifications across all eight files; the
  only additions are the `mod tests;` declarations. `production_lines` in the
  ledger is unchanged for every entry as a result.
- **No test was dropped.** `#[test]`/`#[tokio::test]` counts are conserved per
  file pair — 3,515 total.
- **`cargo test --workspace --lib` is green**, including 3,511 wow-world tests,
  the same count as before the extraction.

Ledger ceilings and the doc table are lowered to the measured values — the first
time in this ratchet's history that they fall rather than rise. The table gains
an "Extracted test sibling" column so the drop reads as a move rather than as
deleted tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


alseif0x:
Added a second commit extending the same move to the eleven remaining files that
still carried a large inline `mod tests`:

| File | Before | After | Sibling |
|---|---:|---:|---:|
| `spell.rs` | 13,736 | 7,559 | 6,185 |
| `update.rs` | 13,439 | 9,735 | 3,712 |
| `map_manager.rs` | 12,935 | 6,607 | 6,336 |
| `spawn_store_loader.rs` | 11,154 | 5,955 | 5,207 |
| `group.rs` | 8,665 | 3,390 | 5,283 |
| `creature.rs` | 6,940 | 3,850 | 3,098 |
| `statements/character.rs` | 6,684 | 3,395 | 3,297 |
| `wow-recastdetour/lib.rs` | 5,664 | 3,194 | 2,478 |
| `lineage.rs` | 4,949 | 3,103 | 1,854 |
| `spell_acquisition.rs` | 4,540 | 2,872 | 1,676 |
| `battle_pet_account.rs` | 4,380 | 3,065 | 1,323 |

Two candidates were left alone rather than forced, because making them fit would
have meant reordering production code instead of moving tests: `packets/misc.rs`
has production code *after* its test module, and `unit_subsystems.rs` has no root
`mod tests` at all. They get their own slice.

Verified identically: production byte-identical in all eleven, `#[test]` counts
conserved per pair, and `cargo test --workspace --lib` green at 35 suites /
7,232 tests / 0 failures. `cargo fmt` re-wrapped some extracted siblings —
dedenting a test body by one level changes the width available to closure and
call arguments — which is whitespace inside test code only.

No ledger ceilings change; none of these files are audited hotspots.

alseif0x:
## Keep the extracted tests inside the hotspot ratchet — correct, and being fixed

This is a real hole and the diagnosis is exact: `validate_hotspot_non_growth`
ratchets only paths curated in the ledger, so capping `session.rs` at 119 test
lines while `session_tests.rs` has no entry leaves 94,042 lines uncapped. The
extraction would have retired the protection it was supposed to leave intact.

### The fix, and why aggregation rather than eight new entries

Both options you name work; aggregating is the one consistent with what this PR
already establishes elsewhere. `#[path]` does not change the module tree — the
child *is* part of the module that mounts it — and the later commit in this
branch teaches the Rust analyzers exactly that, splicing `#[path]` children into
their parent before anything is audited. The Python guard needs the same rule
rather than a second, different one.

So `hotspot_row` now adds each `#[path]` child's counts to the parent that
mounts it, and `hotspot_rows` stops listing mounted children as rows of their
own, since they are already inside their parent's. The consequence is the point:
`session.rs` measures 166,769 / 71,881 / 94,888 again — its original frozen
ceiling — and the ledger entries this PR lowered go back to their pre-extraction
values. Moving lines across the `#[path]` boundary stops changing the number the
ratchet sees, in either direction.

That closes the evasion more generally than adding the eight siblings would: a
ninth extraction tomorrow is covered without anyone remembering to curate it.

### On the rest of the branch

The bridge-specific fix in `bb7cc0bc` is superseded and reverted by the splice
commit — with splicing, the child is not an independent unit at all, so the
production-only filter it widened is correct again. Verified by regenerating:
bridges 71 rows, zero lost, zero new, with the 29 `cfg(test)` rows back under
their parents' paths. The only baseline delta is ten fingerprints, from rustfmt
collapsing an argument list after the dedent and dropping its trailing comma.

The persistence side of the same question is scanning now and will be reported
with numbers rather than asserted.

alseif0x:
Fixed in `1205e6a7`, and one case the suggestion did not cover turned out to
matter.

`hotspot_row` adds each `#[path]` child's counts to the parent that mounts it,
and `hotspot_rows` no longer lists mounted children as rows of their own. Moving
lines across a `#[path]` boundary now changes nothing the ratchet sees, in either
direction, so a ninth extraction is covered without anyone remembering to curate
it.

**The case the first attempt got wrong:** `character_vendor_atomicity_tests.rs`
carries no inner `#![cfg(test)]`, because its parent's declaration supplies it.
Counting the file on its own terms scored it as **169 production lines** and
raised `character.rs` production above its ceiling. Test-only-ness is now read
from the mount as well as from the file.

### Result

Ceilings go back to their pre-extraction values. Exactly one entry rises, and
only in test lines — `character.rs` by 38, the doc headers added to its siblings.
**Production is identical to the frozen baseline in all eight**, which is the
evidence that nothing but tests moved.

| Hotspot | Production | Tests | Total |
|---|---:|---:|---:|
| `session.rs` | 71,881 | 94,165 | 166,046 |
| `map.rs` | 15,245 | 18,273 | 33,518 |
| `character.rs` | 20,200 | 10,691 | 30,891 |
| `loot.rs` | 13,619 | 16,081 | 29,700 |
| `main.rs` | 15,370 | 12,533 | 27,903 |
| `misc.rs` | 7,315 | 11,322 | 18,637 |
| `quest.rs` | 8,255 | 10,172 | 18,427 |
| `player.rs` | 9,265 | 8,891 | 18,156 |

### The persistence side, measured

23,404 → 23,459 rows, **zero lost**. All 24 new keys are in
`void_storage::tests` — one of the two `#[path]` siblings that predate this
branch. Splicing recovers rows their provenance blindness had been dropping all
along, so the inventory grows because the analyzer sees more, not because there
is more code.

`check_architecture.py check` and `session-ownership-check check` both pass:
23,459 rows, 905 semantic groups, 71 bridge rows.

@codex review

alseif0x:
## CI caught a real regression I introduced, fixed in `57edd138`

Seven analyzer tests failed after the previous push. Reporting the diagnosis
rather than just the fix, because one of the seven was not a stale test.

### What broke

The splice ran inside `walk_source_file`, removing the `#[path]` attribute
before the mount graph ever saw it. Two consequences:

- The report began claiming **`0 #[path] modules verified`** while nineteen
  modules were being handled — a false statement about the audit's own coverage.
- `module_aware_registration_scan_uses_logical_mounts_and_rejects_duplicate_owners`
  **stopped failing**. A source mounted under two capability owners passed
  silently, because the child was no longer a mount for the rejection to see.

The second is the actual damage: the audit lost a rejection it exists to make.
The first is only a wrong number, but it is wrong in the direction that hides the
second.

### How the blast radius got bounded

`repository_handler_contract_passes` was the useful one. It failed, but its
output showed the contract itself **passing** — same 478 snapshot rows, same 55
clean sources — with only the `#[path]` count wrong in the report string. That
ruled out the possibility I was most worried about, that module resolution over
the real tree had changed and every measurement reported for this branch was
suspect. It had not.

### The fix

The graph is left exactly as it was — mounts, count, duplicate-owner rejection
intact — and the splice moves to the two places that read a source *in order to
measure it*, which skip `#[path]` children because those already arrive inside
their parent.

That is the shape the Python guard in the previous commit already had. The Rust
side should have matched it from the start instead of inventing a second model
for the same idea.

### Verification

- Analyzer suite: **307 passed, 0 failed** — the seven recovered.
- Exact inventory: **byte-identical** to the frozen one at 23,459 rows, so the
  numbers reported earlier for this branch stand unchanged.
- `check_architecture.py check` and `cargo fmt --all --check`: clean.

### On how it reached CI

I ran the analyzer's full suite before the last splice changes and only the two
splice tests after them. CI found the seven, not me. The suite takes ten minutes
on this machine, which is a reason to run it in the background, not a reason to
skip it.

@codex review

alseif0x:
All four scanner findings fixed. Recording which commit each landed in, since
two of them were holes I opened in the previous round's fix.

| Finding | Resolution |
|---|---|
| Keep the extracted tests inside the hotspot ratchet | `1205e6a7` — a `#[path]` child's counts are charged to the parent that mounts it |
| Preserve path children in the source graph | `57edd138` — the splice moved out of the walker, so the graph keeps its mounts, its count and its duplicate-owner rejection |
| Parse only real path module declarations | `bf6db4a7` — a match is accepted only when it *starts* in code |
| Fix the quoted-path self-test lookup | `8dba5ebe` |
| Parse raw-string path attributes | `bd415456` |
| Match splices against the full mount context | `bd415456` |

Two are worth more than a row.

**The self-test lookup** was mine and it was embarrassing in a useful way: the
test searched for each decoy as `#[path = "name.rs"]`, and the decoy inside a
Rust string literal is written with escaped quotes, so it raised `ValueError` and
took the whole architecture policy self-test down. I had reported the file green
on `check_architecture.py check` exiting 0 — which does not reach that path. CI
runs it through `pr-preflight.sh self-test`. Both are run here now.

**The context key** is the one I would have missed. `(file, logical path)` is not
unique: `#[cfg(not(test))] mod foo;` beside `#[cfg(test)] #[path = "foo.rs"] mod
foo;` gives two contexts with the same name, so filtering by name removed the
production one that no splice represents, and its persistence left the inventory
with nothing said. The key carries the cfg now.

That second fix is also why the inventory check matters here rather than being
ceremony: a key that filtered too much would have dropped real contexts, which is
the same failure the aggregation exists to prevent. It is byte-identical at
**23,459 rows**.

Verification, through both paths this time: `pr-preflight.sh self-test` passes,
analyzer suite 307 passed / 0 failed, architecture guard clean, both formatters
clean.

@codex review

alseif0x:
Fixed in `3b66da74`.

`#[path = "foo\x2ers"]` is the same mount as `foo.rs` to rustc, and the pattern returned the undecoded spelling — so the child was never found and its lines went uncharged, which is the raw-string bypass again wearing a different hat. The plain-literal branch also stopped at the first `"`, so an escaped quote in the name broke the match outright.

Ordinary literals are decoded now (`\xNN`, `\u{..}`, the simple escapes, line continuation); raw strings have no escapes and are left alone. The self-test fixture carries the escaped spelling beside the plain and raw forms.

Verified through both paths: `pr-preflight.sh self-test` and the architecture guard pass.

@codex review

alseif0x:
Fixed in `271adef3`. Two of the three findings were the same problem, so the fix
removes the class rather than the instances.

**The scanner is gone.** An invoked `macro_rules!` generates a real mount that
masking hides, and `# /* keep */ [path = "child.rs"]` is a declaration `syn`
accepts and a pattern does not. Neither is reachable by another pass: finding
these in text means reimplementing a Rust lexer, and the gap list so far has run
comments, escapes, raw strings, char literals versus lifetimes, multi-line
declarations, macro bodies and now token trivia — each one a way to move a
ceiling, with macro expansion out of reach entirely.

`session-ownership-check print-path-modules` emits what `syn` already resolved
and the guard consumes it: **22 mounts, 9.6s**. `rust_code_offsets`, the
raw-string prefix, the macro head, the escape decoder and the declaration
pattern are deleted. The self-test now checks the mapping is non-empty and
consistent, because lexing is no longer this file's job.

**The shebang was a real splice bug** and worth its own note: rustc and rustfmt
accept `#!/usr/bin/env false` atop an external module file, but inside `mod
child { .. }` that line starts an inner attribute and does not parse — so the
reconstructed parent became unparseable and every row it owned silently left the
inventory. Stripped, keeping the newline so later offsets do not shift.

Analyzer suite **308 passed / 0 failed**, architecture guard and
`pr-preflight.sh self-test` clean, exact inventory byte-identical at 23,459.

@codex review

alseif0x:
Fixed in `f2837911`.

The previous fix skipped trivia before deciding whether `#!` opens a shebang, but bounded the search to the first line — and a line comment ends at a newline, so `#!// keep` followed by `[cfg(test)]` is one inner attribute across two lines. The first-line rule deleted the `#!//` and left a stray `[cfg(test)]` that does not parse. The search is no longer bounded, with a test for that spelling.

Analyzer suite **309 passed / 0 failed**, architecture guard, `pr-preflight.sh self-test` and both formatters clean.

@codex review

alseif0x:
Fixed in `1c9d887c`.

Block comments nest and the trivia skipper stopped at the first `*/`, so `#! /* outer /* inner */ end */ [cfg(test)]` still lost its first line. Depth is counted.

Test-only-ness compared the normalized cfg against the literal `cfg(test)`, so `cfg(all(test))` read as production-capable and would have charged a wholly test child against a production ceiling. It is evaluated as a predicate now: `all(..)` is test-only when any term is; `any(..)` and `not(..)` are not, since neither guarantees `test` holds.

22 mounts resolved, 21 test-only. Analyzer suite **309 passed / 0 failed**, guard, preflight and formatter clean.

@codex review

alseif0x:
The ordinary-external-module gap is real and I am not fixing it here: filed as
**#220**.

This PR closed the `#[path]` hole, which is the one the existing extractions
actually use. Covering `mod foo;` means implementing Rust's default path
resolution (crate root vs nested, `foo.rs` vs `foo/mod.rs`, `mod` inside inline
`mod x { .. }` blocks) and re-freezing every ratchet against the new mount set —
a change of the same size as this PR, not an addendum to it. #220 carries the
acceptance criteria, including that converting an existing `#[path]` child to
the ordinary form must leave every hotspot row byte-identical.

Nothing in the tree today uses the ordinary form for an extraction, so the gap
is not currently exploitable; it becomes reachable the moment #150 or the
`session.rs` split starts, which is why #220 is a prerequisite for that work
rather than a nice-to-have.

@codex review

PR #217, opened 2026-08-19
[CI.1] Judge Codex reviews by their findings, not by their boilerplate

### LINKED ISSUE #216: [CI.1] Codex review gate fails clean reviews, blocking every PR
`.github/workflows/codex-review-gate.yml` decides the verdict from the review
*body text*, and `"automated review suggestions"` is in its negative-marker list.
But that phrase is Codex's boilerplate header, emitted whenever it posts a body
at all — including on a review that reports nothing.

Observed on PR #212: review on `7e32308df3` has a body consisting only of the
header plus the "About Codex in GitHub" details block, and **zero inline
comments**. The gate reports "Codex left review feedback for current HEAD;
address it before merge" with nothing to address, so the PR cannot go green and
cannot be merged. Codex's own note in that body says it reacts 👍 when it has no
suggestions, but it demonstrably also posts an empty-bodied review instead.

This blocks any PR whose review lands in that shape, not just #212.

### Fix

Judge a review by whether it carries findings rather than by boilerplate:

- a P0–P3 badge anywhere in the text is a finding and always fails;
- otherwise a **review** is decided by its inline comment count;
- an **issue comment** has no inline comments, so there the header remains the
  only available signal and still counts as feedback.

### Done criteria

- The verdict logic is exercisable outside a CI run, following the repo
  convention of a tool carrying its own self-test subcommand.
- Cases cover both directions: an empty review passes, and a review with inline
  comments or a P-badge still fails.
- PR #212 and #215 reach a green `Codex reviewer verdict`.

alseif0x:
Closes #216

The gate classified a review from its body text, with `"automated review
suggestions"` in the negative-marker list. That phrase is the header Codex prints
whenever it posts a body at all — including on a review that reports nothing.

PR #212's review on `7e32308df3` is a body of nothing but that header and the
About-Codex block, with **zero inline comments**. The gate answered *"Codex left
review feedback for current HEAD; address it before merge"* with nothing to
address, so the PR could not go green and could not be merged. Any PR whose
review lands in that shape is blocked the same way — this is not specific to #212.

### What changes

A review is judged by whether it actually carries findings:

- a P0–P3 badge anywhere in the text is a finding and always fails, since the
  finding is stated in the body;
- otherwise a review is decided by its **inline comment count**;
- an issue comment has no inline comments to count, so there the header stays the
  only available signal and still counts as feedback.

The gate stays strict in the direction that matters. Only the empty-review case
changes verdict — confirmed by running the old classifier and the new one against
the real #212 body: `fail` before, `pass` now.

### Why it moved out of the workflow

A gate that silently blocked a mergeable PR should not be verifiable only by
running CI. The logic is now `tools/ci/codex_review_gate.py --self-test`,
matching how `check_architecture.py` carries its own self-tests. Seven cases
cover both directions:

```
ok   'pass' review carrying only the boilerplate header and no inline comments is clean
ok   'fail' review with the same header but inline comments is feedback
ok   'fail' a P-badge outranks a zero inline count, since the finding is in the body
ok   'fail' issue comment keeps the header as its only signal: no inline comments exist
ok   'pass' issue comment stating no major issues passes
ok   None   unrelated chatter yields no verdict, so the gate keeps waiting
ok   'pass' empty review body with no inline comments is still clean
```

Verified the extraction is faithful: diffing the original heredoc against the new
file, ignoring indentation, shows only the intended edits and nothing else.

The job gains a checkout since the script is no longer inline. It takes the
script from the pull request, as `pull_request` already does for the workflow file
itself — pinning it to the base would be protection in appearance only, because a
branch can edit this workflow either way, and it would also leave the script
missing until this commit is on `3.4.3`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/codex-review-gate.yml
+++ b/.github/workflows/codex-review-gate.yml
@@ -11,6 +11,9 @@ on:
       - ready_for_review
 
 permissions:
+  # actions: read is what lets the gate ask when this run was created, which is
+  # the only server-observed boundary that a thumbs-up can be measured against.
+  actions: read
   contents: read
   issues: read
   pull-requests: read
--- a/.github/workflows/codex-review-gate.yml
+++ b/.github/workflows/codex-review-gate.yml
@@ -11,9 +11,6 @@ on:
       - ready_for_review
 
 permissions:
-  # actions: read is what lets the gate ask when this run was created, which is
-  # the only server-observed boundary that a thumbs-up can be measured against.
-  actions: read
   contents: read
   issues: read
   pull-requests: read
--- a/.github/workflows/codex-review-gate.yml
+++ b/.github/workflows/codex-review-gate.yml
@@ -25,6 +25,18 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 35
     steps:
+      # The verdict logic moved out of this file into tools/ci, so the job needs
+      # the repository. It comes from the pull request like the workflow file
+      # itself does on `pull_request`, which is the same trust model as before:
+      # a branch could always edit this workflow to weaken its own gate, so
+      # pinning the script to the base would be protection in appearance only.
+      - name: Check out pull request
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          sparse-checkout: tools/ci
+          sparse-checkout-cone-mode: false
+
       - name: Require clean Codex review on current HEAD
         env:
           GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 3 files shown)


alseif0x:
Second commit: the gate has a **second** way of failing a clean review, and this
PR hit it itself.

Codex's own note reads *"If Codex has suggestions, it will comment; otherwise it
will react with a thumbs up"*. On a PR it is happy with, Codex writes nothing at
all. This PR is exactly that — a 👍 reaction, **zero reviews, zero comments**,
every other check green — and the gate polled for 30 minutes and failed with
*"Timed out waiting for a clean Codex review"*.

So between the two commits this PR covers both ways Codex says "nothing to say":

| Codex signal | Old gate | Now |
|---|---|---|
| Review with boilerplate header, no inline comments | `fail` — "address it before merge", nothing to address | `pass` |
| 👍 reaction, no review at all | times out after 30 min → `fail` | `pass` |
| Review with inline comments | `fail` | `fail` |
| P0–P3 badge in the body | `fail` | `fail` |

A reaction carries no commit reference, so the only thing tying it to a revision
is that Codex reacts *after* reviewing. A thumbs-up older than the HEAD commit is
ignored — otherwise the check would go green on unreviewed code the moment anyone
pushed a fix on top of an approved commit. A missing timestamp on either side
fails closed, and the reaction is consulted only when nothing was written about
this HEAD, so a real finding always outranks it.

That rule applies to this PR too: the 👍 above predates `df58f58a`, so the gate
will correctly keep waiting until Codex weighs in on the new HEAD.

Self-test is now 11 cases, four covering the reaction rule in both directions.

alseif0x:
The P1 is real and is fixed in `e782c05a`.

Comparing the reaction against the HEAD commit's **committer date** does not bind
it to that revision: committer dates are author-controlled and record when a
commit was *written*, not when it reached GitHub. The sequence you describe —
commit created locally at 10:00, thumbs-up on the old head at 10:30, push at
11:00 — passes under that rule, and a back-dated force-push does the same.

The boundary is now the **earliest check suite for the HEAD SHA**. GitHub creates
a suite when a SHA arrives, so it is a server-observed lower bound on when that
revision became reviewable, and nothing the author writes can move it.

One gap remains, and I would rather state it than imply it is closed: if Codex
began reviewing the previous HEAD, a push landed mid-review, and Codex then
reacted, the reaction postdates this SHA while describing the previous one. No
field in the reactions API distinguishes those — a reaction carries no commit
reference, unlike a review, whose body states the SHA it read. Codex re-reviews
on every synchronize, so the next signal settles it, and because a finding always
outranks a reaction the failure mode is a *delayed pass*, never a suppressed
finding. Closing it properly needs SHA-stamped evidence that a bare thumbs-up
does not carry.

Self-test is 13 cases now, including the back-dated-commit sequence itself and a
reaction landing exactly on the boundary.

Worth noting for this PR's own history: the previous run is the gate working
correctly for the first time. It read your review, found inline comments, and
failed — which is exactly the behaviour the first commit was meant to restore.

@codex review

alseif0x:
### The in-flight review gap: removed, not patched

You are right, and my own defence of it was wrong. I wrote that the worst case
was *"a delayed pass, never a suppressed finding"*. It is not. If Codex is
reviewing SHA A when a push makes B the head, its eventual thumbs-up about A
lands after this run began, the gate passes B, and B can be merged before Codex's
review of B ever arrives. A finding is **lost**, not delayed.

No field can fix this. A reaction carries no commit reference; a review's body
states `Reviewed commit: <sha>`. Three boundaries were tried — the committer date,
the SHA's earliest check suite, this run's `created_at` — and each failed in its
own way, because none of them binds a *reaction* to a *revision*. So `1920f4fc`
removes the mechanism instead of patching it a fourth time, along with
`gh_json_object`, `actions: read`, and the six reaction self-tests.

Clean PRs still pass on SHA-stamped evidence: Codex does post empty-bodied
reviews that name the commit — PR #212's review on `7e32308df3` is exactly that —
and reading those is the first commit in this branch, untouched. When Codex only
reacts, the gate now times out and says why, and `@codex review` produces a
review that states the SHA it read.

### Running from the protected base: the premise does not hold

This one I have to push back on, with evidence from this PR's own runs.

The claim is that "the workflow definition itself is selected from the protected
base branch". For `pull_request`, it is not — GitHub takes the workflow file from
the PR's merge ref, which contains the head's version. Run
[32249906543](https://github.com/alseif0x/rustycore/actions/runs/32249906543)
executed the step **"Check out pull request"**, a step that exists only in this
branch and not on `3.4.3`. The definition came from the head.

So a branch that wanted to neuter this gate would not need to touch the script:
it could rewrite the workflow's own `run:` step, which the same event model
already permits. Pinning the script to the base would move the attack one file to
the left, not prevent it — and it would also leave the script missing until this
commit is on `3.4.3`, since `tools/ci/` does not exist there yet.

There is a weaker version of the argument I do accept: a change to a workflow
file is more conspicuous in review than a change to a script it calls. That is a
review-visibility benefit, not a security boundary, and it is not what the
finding claims. If you want the base-pinned form regardless, say so and I will
implement it after this lands, when the bootstrap problem is gone.

@codex review

alseif0x:
## Thread resolutions

Four of the five findings were real and are fixed; the fifth is rebutted with
evidence rather than implemented. Recorded here so the record does not depend on
reading the diff.

| Finding | Resolution |
|---|---|
| Boilerplate header read as feedback | Fixed `eb4ff1dd` — a review is judged by its inline comment count and P-badges, not by the header Codex prints whenever it posts a body |
| Thumbs-up bound to the committer date | Fixed `e782c05a` — committer dates are author-controlled; moved to a server-observed boundary |
| Check-suites boundary needed `checks: read` and belonged to the SHA, not this PR | Fixed `d061164c` — moved to this run's `created_at` with `actions: read` |
| In-flight prior-head review could pass an unreviewed HEAD | Fixed `1920f4fc` — **mechanism removed**. You were right and my defence of it was wrong: I claimed the worst case was a delayed pass, but a finding can be lost outright. No field binds a reaction to a revision, so the reaction path is gone rather than patched a fourth time |
| Run the gate implementation from the protected base | **Won't fix — premise does not hold** |

### On the last one

The claim is that "the workflow definition itself is selected from the protected
base branch". For `pull_request` it is not. Run
[32249906543](https://github.com/alseif0x/rustycore/actions/runs/32249906543)
executed the step **"Check out pull request"** — a step that exists only on this
branch and not on `3.4.3`. GitHub took the workflow from the head.

So a branch wanting to neuter this gate never needs to touch the script: it can
rewrite the workflow's own `run:` step under the same event model. Pinning the
script to the base moves the attack one file to the left, to the `.yml`, and does
not prevent it. It would also leave the script missing until this commit is on
`3.4.3`, since `tools/ci/` does not exist there.

The weaker form of the argument is fair and I am recording it rather than
dismissing it: a change to a workflow file is more conspicuous in review than a
change to a script it calls. That is review visibility, not a security boundary,
and it is not what the finding claims. If the base-pinned form is wanted anyway,
it is a clean follow-up once this is on `3.4.3` and the bootstrap problem is gone.

### Gate status

`Codex reviewer verdict` is **green on `1920f4fc`** — the first time it has passed
on this branch, and it passed by reading a real review rather than by being
bypassed. The two bugs this PR fixes were the only two ways a clean review reached
it, which is why every recent PR was stuck.

Issue #252, opened 2026-08-21
[ARCH.RUNTIME.PLAYER] Retire the temporary PlayerBroadcastInfo gameplay mirror


alseif0x:
Parent: #133. Depends on #28, #138, #140 and #189.

## Problem

After the session directory is opaque and relocated, `PlayerBroadcastInfo` still carries a broad 80-field copy of Player, Map, visibility, loot and Session state. Keeping that compatibility mirror indefinitely would preserve a second authority and make the directory a gameplay query-anything service.

## Goal

Retire the temporary `PlayerBroadcastInfo` gameplay mirror field family by field family, replacing it with minimum owned/versioned queries from the canonical Player/Map owners and opaque generation-checked Session addressing. This is one bounded strangler cut, not a Player rewrite.

## Scope

- Follow the exact field ledger produced by #196; no field may be added or left unassigned.
- Remove mailbox/delivery and durable-loot fields only after #140/#189 own those capabilities.
- Replace Player-local identity, vitals, progression, inventory, quest, reputation, social and presentation copies with minimum immutable Player-owned results.
- Replace position/map/instance/transport/liquid and spatial visibility copies with Map traversal plus Player-owned visibility membership.
- Preserve generation fencing, C++-observable packet/order semantics and existing canonical writers.
- Remove `PlayerBroadcastInfo` when the final field retires; do not create another mega-projection, mutable mirror or `Arc<Mutex<Player>>`.
- Split implementation commits by field family, but keep one directory contract and one closing PR.

## Acceptance

- `PlayerBroadcastInfo` and its exact-field architecture baseline are gone.
- The session directory retains only opaque identity/incarnation/addressing entries and named bounded operations.
- Each gameplay value has one canonical owner/writer; no Session or directory mirror can mutate it.
- Cross-session delivery, replacement/unregister, movement/visibility, group/social, loot, quest and inspect suites remain green.
- Focused C++/capture evidence is required only where a cut changes observable behavior; otherwise the cut must be behavior-preserving.
- Validation uses `./tools/local-harness.sh final origin/3.4.3` plus focused affected suites.

Refs #133, #138, #140, #189 and #196.

alseif0x:
## Measured at `3622e703`: the projection is far more encapsulated than its field count suggests

A naive count says 72 fields and ~5,100 reads across the workspace. That number is an upper
bound and badly inflated: it matches `.position`, `.map_id`, `.account_id` on *any* struct with
[...]

Issue #223, opened 2026-08-21
[ARCH.PLAN] Align the modular roadmap, physical metrics and local-first policy


alseif0x:
Parent: #133. This synchronization work may run in parallel and does not block production splits.

## Problem

The checked-in architecture ledger still describes the old serial DAG, omits live follow-ups and encodes a validation workflow superseded by #222. The current hotspot ratchet also aggregates child modules, so it cannot measure the physical file/module structure requested by the project.

## Goal

Make the repository documentation and offline plan match the 2026-08-21 GitHub architecture reset without creating another expensive analyzer.

## Scope

- Update the architecture roadmap and issue ledger to the current four lanes:
  1. physical modules;
  2. encapsulation and ownership;
  3. runtime/persistence authority;
  4. stable crate and extension seams.
- Record every live architecture issue exactly once and remove obsolete serial dependencies.
- Treat #153 as the closure audit, not as the gate for already-known physical work.
- Publish two independent views:
  - physical lines per real file, with production/tests/generated separated;
  - logical owner size including private descendants.
- Document soft review signals: inspect files around 2,000 productive lines and require a split plan or explicit exception above 4,000. These are health signals, not a daily blocking gate.
- Align architecture instructions and issue templates with docs/operations/local-first-development.md.
- Keep exhaustive architecture/persistence audits available on demand and scheduled, outside the ordinary first-party path.
- Prefer a simple existing report or a small script over extending the 21k-line persistence analyzer.

## Acceptance

- GitHub #133, the checked-in ledger and ownership documentation describe the same lanes and live issues.
- Ordinary mod declarations reduce the physical-file metric while logical ownership remains honest.
- The plan no longer requires #220's child-charging change to recognize a successful physical split.
- Validation follows docs/operations/local-first-development.md: tools/local-harness.sh final plus the focused evidence required above. Remote checks follow #222; full preflight and model review are opt-in.
- External contributors retain the remote checks defined by #222.
- No gameplay, packet, persistence or runtime behavior changes.
- The implementation remains a small documentation/reporting PR and does not block #136, #139, #152 or the new physical splits.

Validation: tools/local-harness.sh final origin/3.4.3 plus focused checks for any reporting script changed.

Issue #199, opened 2026-08-11
[ARCH.2.5d] Separate group persistence/publication and close GroupRegistry storage


alseif0x:
Parent: #133. Depends on #198. Uses focused Group durability tests. Followed by #195 and the mechanical move #137.

## Defect

After semantic operations exist, persistence materialization, session-command publication, fixtures and public compatibility helpers can still couple Group to networking/database implementation and keep its backing map accessible.

## Goal

Finish the Group boundary: the owner exposes typed commands/queries/outcomes, while application adapters perform ordered persistence and delivery after mutation.

## Scope

- Introduce typed Group persistence intents/outcomes preserving Group transaction/publication behavior with focused tests.
- Move packet/session-command fanout orchestration out of locked Group mutation.
- Migrate compatibility fixtures to test-only builders or invariant-preserving public operations.
- Make GroupRegistry/PendingInvites storage private and remove raw aliases, iterators, entry guards and generic mutation closures.
- Ensure the Group core imports no `wow-network`, SQLx, pool or packet-encoder type, enabling the later move to `wow-social`.
- Do not perform physical relocation; #137 is the next mechanical PR.

## Acceptance

- Zero direct backing-storage access outside the owner module.
- Focused Group transaction/commit/publication fixtures remain exact.
- Group concurrency and packet-order suites pass.
- No database/network type, lock-across-I/O or service locator leaks into core API.
- #195 can route outcomes through session addressing and #137 can move the core without behavior change.
- Validation follows docs/operations/local-first-development.md: tools/local-harness.sh final plus the focused evidence required above. Remote checks follow #222; full preflight and model review are opt-in.

Refs #133, #137, #151, #195 and #198.

PR #257, opened 2026-08-21
[ARCH.2.5e] Move the encapsulated Group owner into wow-social

### LINKED ISSUE #137: [ARCH.2.5e] Move the encapsulated Group owner into wow-social
Parent: #133. Depends on #195 and #199. The mailbox move is owned by #140.

## Defect

After its invariants and adapters are separated, the Group owner remains physically under `wow-network`, making transport appear to own social gameplay and retaining an invalid dependency direction.

## Goal

Mechanically relocate the already-opaque Group/PendingInvites core and colocated invariant tests into the existing `wow-social` crate. This PR changes physical ownership only.

## Scope

- Move Group state, commands, queries, outcomes and invariant tests to coherent private submodules under `wow-social::group`.
- Preserve one atomic authority and all APIs established by #151/#197/#198/#199.
- Keep packet encoding, session addressing/delivery and persistence adapters outside `wow-social`.
- Remove obsolete `wow-network` re-exports and high-level dependencies when the final consumer disappears.
- No Group behavior, packet bytes/order, DB order, runtime clock or Player ownership change.

## Target submodules

- `group/model.rs`: IDs and owned state.
- `group/invites.rs`: invite lifecycle.
- `group/membership.rs`: create/join/leave/roles/subgroups.
- `group/settings.rs`: loot/difficulty/target icons.
- `group/ready_check.rs`: ready-check state machine.
- `group/outcome.rs`: ordered facts/intents, with no transport/SQL types.

Submodules may be combined only when the ownership inventory proves cohesion; their public surface remains the narrow `GroupService`/store boundary.

## Acceptance

- Group tests run from `wow-social` unchanged and all focused persistence/publication behavior remains exact.
- `wow-network` owns no Group/PendingInvites gameplay state.
- `wow-social` has no dependency on `wow-network`, `wow-session`, SQLx or packet encoding.
- No compatibility re-export remains without a concrete removal issue.
- Validation follows docs/operations/local-first-development.md: tools/local-harness.sh final plus the focused evidence required above. Remote checks follow #222; full preflight and model review are opt-in.

Refs #110, #133, #140, #151 and #195–#199.

alseif0x:
Closes #137

## What this is

The atomic Group authority leaves raw networking. The 5,114-line `wow_network::group_registry` becomes `wow_social::group`, split into the private submodules the issue names, with `group/mod.rs` as the only public surface:

| Submodule | Contents |
|---|---|
| `model` | identity, ID allocators, owned `GroupInfo`, opaque `GroupRegistry`, DB-row loaders |
| `invites` | invite lifecycle + `PendingInvites` |
| `membership` | join/leave/roles/flags/subgroups |
| `settings` | loot method + looter rotation, difficulty, target icons, raid markers |
| `ready_check` | ready-check state machine |
| `outcome` | ordered authority facts and persistence/publication intents — no transport or SQL types |

All **80** Group invariant tests move with the owner and pass unchanged. Following the #214 extraction convention their only textual change is dedenting by one level.

The split was generated from an **exact line-range partition** verified to leave no source line unassigned, not by hand-copying. No Group behaviour, packet bytes or order, DB order, runtime clock or Player ownership changes.

## The one design cut the move forced

`Group::LoadGroupFromDB` needs C++'s loaded-difficulty validation, but `wow-social` is `domain-runtime` and `wow-data` is `adapter-platform` — the direct edge would be **upward and illegal**.

So `wow-social` states the requirement as the narrow port `GroupDifficultyValidatorLikeCpp`, preserving C++'s exact fallback contract (an unusable id falls back to the normal difficulty of its instance kind). The composition root binds it with the explicit borrowing adapter `GroupDifficultyStorePortLikeCpp` — a newtype rather than a blanket impl, because both the trait and `wow_data::DifficultyStore` are foreign to `world-server`.

No service locator, event bus, `Arc<Mutex<_>>` or second mirror. The Group tests keep exercising the **real DB2 store** through a development-only `wow-data` edge (explicitly outside the production boundary per the policy) rather than a weaker hand-written stand-in.

## Dependency direction

`wow-network` now owns no Group state: its `group_registry` module and 48 crate-root re-exports are gone.

Two exception entries, both mailbox-scoped and both owned by #140:

- **new** `wow-network → wow-social` — the Session mailbox still names `GroupDifficultyKindLikeCpp` in `ApplyGroupDifficultyLikeCppCommand`. It is the only Group symbol left in `wow-network`.
- **re-owned** `wow-network → wow-data` — this exception previously named **#137** as its remover. Two mailbox payloads still carry `wow_data::quest::QuestTemplate`, so #137 cannot remove it; it is re-owned by #140 rather than left pointing at a completed slice.

`wow-social` declares `dashmap` in the external allowlist (the moved owner's existing concurrent storage). Workspace edges: 106 → 109.

## Guards follow the owner (never weakened)

- `session-ownership-check` now scans `wow-social`. **Without this, `GroupDifficultyKindLikeCpp` would have silently left the pinned `SessionCommand` payload inventory** — the first `--syntax-only` run showed the obsolete payload entry with no replacement.
- The anti-glob registry guard gains `wow_social`/`group`, with a **new adversarial test** proving `use wow_social::group::*;`, `use wow_social::*;` and `use crate::group::invites::*;` all fail closed while an unrelated module glob stays allowed.
- The relocated owner enters the hotspot table with its **own non-growth ceiling** (2,783 production / 2,511 test) instead of landing untracked.

## Reviewed baseline delta

Inspected row by row, not regenerated to go green:

| Surface | Delta |
|---|---|
| `registry_accesses` | 572 → **570**: 17 rows relocate one-for-one; the two obsolete `wow-network` `GroupRegistry`/`PendingInvites` re-export rows disappear |
| `world_session.fields` | **738 unchanged**; one field's type path follows the move |
| `session_command` / payload types | **37 / 42 unchanged**; one payload's module follows the move |
| `generated_surface_inputs` | **44 unchanged**; same derive remapped |
| `bridge_accesses` | **71 rows, identities identical** |
| `PlayerBroadcastInfo` | **80 fields**, untouched |

## Finding reported, not hidden

While syncing the offline issue ledger against GitHub I found the snapshot lists **15 stale states** (#137–#139, #150–#152, #192–#199, #223 are all CLOSED on GitHub but `open` in the ledger). Syncing them exposes **four pre-existing DAG violations** the stale snapshot was masking:

```
closed #192 depends on open #188
closed #193 depends on open #187
closed #194 depends on open #188
closed #199 depends on open #187
```

Those are four other merged slices declaring still-open prerequisites. Resolving that is not a mechanical Group move, so **I reverted the sync rather than paper over it or silently expand this PR's scope**. It needs its own issue.

## Validation

| Command | Result |
|---|---|
| `cargo check --workspace --all-targets` | clean, **no new warnings** |
| `cargo test -p wow-social --lib` | **80/80** — the moved invariants, unchanged |
| `cargo test -p wow-network --lib` | **45/45** — exactly 125 − 80, confirming a clean split |
| `cargo test -p wow-world --lib group` | 240/240 |
| `handler-contract-check registry_inventory` | 8/8 incl. the new adversarial test |
| `session-ownership-check -- check --syntax-only` | PASS (570 exact direct-registry rows) |
| `check_architecture.py check` | PASS (34 packages, 109 edges, 8 logical owners, 20 exceptions) |
| `check_architecture.py self-test` | PASS (20 fixtures) |
| `cargo fmt --all -- --check`, `git diff --check` | clean |
| `./tools/local-harness.sh final origin/3.4.3` | **passed** |

Per the local-first policy for `alseif0x`, hosted jobs intentionally skip; the evidence above is the gate. No capture-diff or runtime QA: no bytes, metadata, connection choice or observable ordering changed.

## Remaining boundaries

`SessionCommand` and its payloads, the durable rails, physical channels, socket fences and `InstanceLink` stay in `wow-network` (#140/#189). `PlayerBroadcastInfo` remains the temporary 80-field mirror (#252).


alseif0x:
Follow-up for the ledger drift reported in this PR's description: **#258**.

Correction to that section: the accurate count is **two** violations, not four. `#193 → #187` and `#199 → #187` were pure snapshot staleness — #187 is closed on GitHub. The two real ones are `#192 → #188` and `#194 → #188`, both against a genuinely open #188. Full measurement against `3.4.3` @ `5b3f85ea`: 16 stale ledger states, 2 real prerequisite violations, 0 dependency-policy exceptions owned by a completed issue.

Issue #200, opened 2026-08-11
[ARCH.DB.4] Earn wow-persistence with the Player lifecycle port


alseif0x:
Parent: #169 and #133. Depends on #184 and the focused contract #187; #186 is already complete.

## Goal

Create wow-persistence by moving one real production vertical: Player login/load/save/logout persistence. The crate does not exist before the vertical and is not a placeholder.

## Ownership

- wow-persistence owns SQLx-free PlayerLifecyclePort inputs, outputs, plans, fences and normalized outcomes.
- wow-database implements the MariaDB/SQLx adapter and remains the only concrete pool/row/transaction owner.
- session/lifecycle owns orchestration and publication through the port.
- world-server composes the adapter.

## Scope

- Define the minimum lifecycle capability from the #187 contract.
- Move concrete SQLx/statement/row handling behind the wow-database adapter.
- Preserve load/save/logout statement order, transaction affinity, commit/rollback/unknown classification, retry/fence and publication.
- Remove the migrated concrete-access exceptions and dependencies from Session/application code.
- Add failure/conformance tests against the real adapter.
- Do not add a generic repository, raw query method, cross-database transaction abstraction or PostgreSQL claim.
- Do not migrate inventory, loot, Group, quest or unrelated loaders in this PR.

## Acceptance

- wow-persistence has zero SQLx, pool, row, raw SQL, packet, network or world dependency.
- Session lifecycle depends on the typed capability and no longer imports the migrated concrete DB surface.
- MariaDB adapter passes the #187 positive/rollback/unknown-order contract.
- Startup rejects a missing/incompatible adapter before accepting sessions.
- No publication occurs before the durable outcome permits it.
- The crate/API exists because production uses it in the same PR.

Validation: tools/local-harness.sh final origin/3.4.3 plus focused lifecycle/MariaDB contract tests. Exhaustive database audits remain opt-in.

alseif0x:
## Scoping after #184 (all prerequisites now closed)

#184, #186 and #187 are closed, so this is unblocked. Measurements of the actual vertical
on `3.4.3` (`5364681c`), so the design decision is made against real code rather than the
issue text.

### Where the port boundary actually falls

`save_current_player_to_db_like_cpp` (`session/mod.rs`, 128 lines) is mostly *orchestration*
that must stay in `session/lifecycle`: the money-admission fence, the durable-loot drain,
the indeterminate-state checks, `kick`, and the quest-objective drain. The concrete
[...]

PR #290, opened 2026-08-22
[ARCH.DB.4b] Move SaveTutorialsData behind the lifecycle port

### LINKED ISSUE #286: [ARCH.DB.4b] Move the Player character-save statement plan behind the lifecycle port
Split out of #200, which landed 200a (the offline-mark vertical) and created
`wow-persistence`. The seam now exists; this moves the largest workflow onto it.

## What moves

`current_player_save_to_db_statement_plan_like_cpp` (`crates/wow-world/src/session/mod.rs`,
**248 lines**, **53 distinct `self.` reads** — level, xp, zone, position, talents,
currencies and more) builds a `Vec<PreparedStatement>`, and
`save_current_player_to_db_like_cpp` commits it:

```rust
let mut tx = SqlTransaction::new();
for statement in std::mem::take(&mut plan.statements) { tx.append(statement); }
let commit_result = tx.commit_with_outcome_like_cpp(char_db.pool()).await;
```

`PreparedStatement` is a `wow-database` type, so the plan cannot cross the port as it
stands. Either the builder moves into the adapter behind an SQLx-free request, or the
Session keeps importing the concrete surface — and the second fails #200's own acceptance.

## The actual work

Enumerate the ~50 represented values the plan reads, field by field, **against the frozen
fixture** `crates/wow-world/tests/fixtures/player-lifecycle-contract.json` (#187) rather
than by reading the Rust and guessing. That enumeration is the deliverable; the mechanical
move follows from it.

## What must not change

- Statement order inside the one CharacterDatabase transaction.
- The three-way commit classification. Unlike the offline marks, this path **can** produce
  a genuine `Unknown`: `PersistenceOutcomeLikeCpp::Unknown` already exists for it, and the
  contract requires it to fence further mutation and force a relogin rather than be read
  as success or rollback.
- Everything around the commit stays orchestration in `session/lifecycle`: the
  money-admission fence, the durable-loot drain, the indeterminate-state checks, `kick`,
  and the quest-objective drain. Only the plan and its execution move.

## Done criteria

- [ ] `wow-persistence` gains an SQLx-free save request; no `PreparedStatement`, pool or
      row appears in it.
- [ ] `session/lifecycle` no longer imports the concrete save surface.
- [ ] Contract tests cover committed, definitely-rolled-back and unknown outcomes, and
      prove dirty state is published only after a successful commit.
- [ ] Statement order verified unchanged against the #187 fixture.

Refs #133, #169, #187, #200.

alseif0x:
First step of #286 — and deliberately not the obvious one. **This PR does not close #286.**

## Why not the 248-line plan

Measured against HEAD, the plan builder is a façade: 34 `self.` reads, **twelve of them
other statement builders** spread across the Session (spells, skills, talents, glyphs,
action buttons, cooldowns, charges, void storage, CUF profiles, equipment sets, instance
restrictions, tutorials, played time). Moving it means moving thirteen builders and
enumerating the data behind every collection.

The tempting shortcut — give the port a statement identifier plus typed params and let the
sub-builders emit that — is precisely what #200 forbids ("no generic repository, raw query
method"), and it would freeze today's statement decomposition into the port's API, making
the real semantic move harder afterwards. See the [scoping
comment](https://github.com/alseif0x/rustycore/issues/286#issuecomment-5381675264).

So this takes the one piece that is genuinely independent.

## What moved

`WorldSession::save_tutorials_data_like_cpp` is C++ `SaveTutorialsData`: its own statement,
its own transaction, **not** part of the character save. It now goes through
`PlayerLifecyclePortLikeCpp::save_tutorials_like_cpp` with an SQLx-free
`PlayerTutorialsSaveLikeCpp` — the flags in index order, the account id, and
`already_persisted` for the INSERT/UPDATE choice, which is data rather than SQL.

The Session keeps what it should: the changed/coherent gating before the call and the state
updates after it.

## One owner for the column order

The full-save plan still appends this same row to its own transaction, so the row shape has
two callers. Rather than duplicate it — two copies of a column order are free to drift, and
drift here writes the wrong tutorial into the wrong column —
`build_tutorials_save_statement_like_cpp` lives in `wow-database` and **both** callers use
it. The Session's builder is now a delegation, and the second caller disappears when the
plan itself moves.

## Dirty state on non-applied outcomes

Neither `Failed` nor `Unknown` clears `tutorials_changed`, and neither records an
unconfirmed INSERT as persisted: after an unknown outcome the row may or may not exist, so
a later save must not assume an UPDATE will match. A test pins both classes.

## Tests

5 new — flags and INSERT/UPDATE choice reaching the port, update on a later save, unchanged
and incoherently-loaded data not written, and the two non-applied classes keeping state
dirty — plus the 5 existing tutorial statement tests, which still pass against the
delegating builder.

## Validation

- `cargo test -p wow-world --lib tests::lifecycle_persistence::` → 10 passed.
- `cargo test -p wow-persistence` → 2 passed.
- `session-ownership-check check` (exhaustive) → PASS, 11929 production + 10524
  test-fixture rows, 932 semantic groups.
- `check_architecture.py check` → PASS.
- `./tools/pr-preflight.sh quick origin/3.4.3` → **exit 0**.

## Inventories

Persistence rows 22443 → 22453, with the migration visible as `session/mod.rs` 1754 → 1734
and the adapter 39 → 69; 2 adapter workflows annotated, 1 obsolete dropped; session hotspot
to measured rows.

Part of #286. Refs #133, #187, #200.


alseif0x:
Correction: `save_tutorials_data_like_cpp` has no production caller — its only callers are the tests this PR added. The method is one of nineteen unreachable standalone save paths in `session/mod.rs`; tutorials are actually persisted through the aggregate plan. This PR's body described it as a production write, which was wrong.

What the PR does still hold up: the SQLx-free `PlayerTutorialsSaveLikeCpp` shape, and moving `build_tutorials_save_statement_like_cpp` into `wow-database` so the *live* plan path and this one share one column order instead of two copies. Tracked in https://github.com/alseif0x/rustycore/issues/291.

Issue #297, opened 2026-08-22
[ARCH.P4.1] Promote the Session kernel to a wow-session crate


alseif0x:
Parent: #133, lane **P4 — Earned crates, not placeholders**. Opened because P4 was the only
lane in the epic with no issue numbers, and #153 forbids absorbing known work
("No known physical split or semantic cut is postponed to this issue"), so #153 cannot
close honestly while this is unassigned.

## The P4 rule

> Promote the kernel to `wow-session` only when it compiles without `wow-world`, gameplay,
> databases, or catalogs.

The crate is **earned, not declared** — the same rule `wow-persistence` followed in #200a,
where the crate appeared in the PR that made production use it.

## Where it stands at HEAD (`2ca7b246`)

`crates/wow-world/src/session/mod.rs` is **74,567 production lines** under one logical
owner — roughly 17× the epic's own hotspot ceiling — and it is a `mod.rs`, which the epic's
principle 3 says must not carry implementation. The physical lane (P1) is fully merged and
moved lines without reducing them; that was encapsulation, not ownership transfer, and this
issue is the transfer.

What already lives in reviewed private submodules and is candidate kernel:

| module | landed by |
|---|---|
| `session/admission.rs`, `session/dispatch.rs` | #152 |
| `session/directory.rs` | #138 |
| `session/mailbox/` | #140 |
| `session/connection.rs` | #182 |
| `session/driver/` | #183 |
| `session/lifecycle/` | #184 |

Those six are the plausible kernel: connection routing, the phase driver, admission,
dispatch, the mailbox and the lifecycle. What blocks extraction today is that every one of
them is an `impl WorldSession` block, and `WorldSession` holds 729 production fields
including catalogs, database handles and gameplay state.

## What this issue must establish first

Do **not** start by creating the crate. Start by measuring, because the answer decides
whether this is one move or a decade:

- [ ] For each of the six candidate modules, which `WorldSession` fields it actually reads.
      `tools/architecture/runtime-ownership-ledger.json` already partitions all 729 fields
      into responsibility families — use it rather than re-deriving.
- [ ] Which of those fields belong to the `immutable_catalogs_configuration_and_services`
      family (186 fields) and are therefore *not* kernel.
- [ ] Whether a kernel type can own the transport/driver/admission fields while gameplay
      state stays on a separate type, or whether the split needs `Player` to land first
      (#252, blocked on #28).

## Done criteria

- [ ] A measured answer to the question above, published on this issue before any code moves.
[...]

PR #376, opened 2026-08-25
Retire the first three PlayerBroadcastInfo field families (#252)

### LINKED ISSUE #252: [ARCH.RUNTIME.PLAYER] Retire the temporary PlayerBroadcastInfo gameplay mirror

alseif0x:
Part 1 of #252. Three field families retire from the `PlayerBroadcastInfo`
gameplay mirror: **72 -> 60 fields**. This does not close the issue, and the
reason is a measurement, not a shortcut — see "Why this issue stays open".

## The finding that shaped this PR

The mirror existed because a remote session appeared unable to reach another
player's canonical `Player`. It always could:
`canonical_player_snapshot_like_cpp` already resolves the value out of the
shared canonical `MapManager`. The data was never unreachable — only
*unaddressable*, because that accessor is bound to `self` and therefore answers
for exactly one GUID. A new placement-addressed accessor removes that
limitation, and with it the reason for the copy.

Resolving all 72 fields through their producer splits them in two:

- **Canonical-backed (~19).** The canonical `Player` owns the value and the
  mirror is a copy. These retire outright. Twelve did, here.
- **Session-owned (~53).** The direction is reversed: `WorldSession` holds the
  copy (`player_health_like_cpp`, `player_level_like_cpp`, the `represented_*`
  family) and *writes it into* the canonical `Player` in
  `canonical_player_entity_snapshot_for_map_like_cpp`.

## What landed

| family | fields | reader |
|---|---|---|
| honor block | 7 | `inspect_snapshot` |
| unit flags + contested PvP + reputation flags/forced factions | 4 | `legacy_aggro_candidates` |
| reputation standings | 1 | `legacy_aggro_candidates`, `quest_sharing_snapshot` |

Each cut removed a staleness window, not just a copy. Every one of these values
was written at registration and refreshed again on each
`sync_player_registry_state_like_cpp` pass, so a read landing between the two
saw the old value. Reading the owner makes that unrepresentable.

## Lock discipline

No path holds a registry shard guard and the canonical mutex at once. Single-GUID
resolvers copy the placement out of the entry and drop the `DashMap` guard before
calling in. The aggro family needed no new lock at all: its only caller already
runs a canonical pass over the collected candidates, taking the map lock once for
the whole batch, so the four reads moved into that existing pass.

## Why this issue stays open

The Acceptance has two halves. *"`PlayerBroadcastInfo` is gone"* is mechanically
reachable for the session-owned fields too, by pointing the builders at the
canonical copy the session already maintains. *"Each gameplay value has one
canonical owner/writer; no Session or directory mirror can mutate it"* is not:
the session would still own and write the value.

Doing the first without the second would trade an explicit projection for an
implicit dependence on a session-written copy, delete the visible evidence of the
second authority, and let the issue be marked done with the problem intact. **Category B
is blocked on Player state leaving `WorldSession`, which is #297's kernel
promotion.** The dependency is recorded on #252 rather than worked around.

## Evidence

- `wow-world --lib` 3579/0 · `world-server` 453/0 · `wow-map --lib` 688/0
- `cargo fmt --all --check` clean
- `./tools/validation-v2 final --base origin/3.4.3` — manifest `passed`
- `check_architecture.py check` + `self-test` green
- exhaustive `session-ownership-check -- check` — `PASS`, and it now counts
  **60 PlayerBroadcastInfo fields**
- The exhaustive `session-ownership-check -- check` caught two real things, both
  fixed rather than waived:
  1. Retiring six now-dead `WorldSession` accessors left their exact impl-item
     baseline stale. Regenerated via `print-baseline` into a temp file and
     reviewed before installing — the delta is exactly those six entries removed,
     nothing added — then only that array was installed, not the whole file.
     `check --syntax-only` reports PASS at 3323 exact associated items.
  2. The new quest fixture took `&WorldSession` to read three values. That type
     carries database handles, so it registered 21 untracked direct persistence
     accesses. The fixture now takes `account_id`, the GUID and the position
     instead, which removes the accesses at the source rather than registering
     them.
- Baselines moved with the code: mirror syntax baseline and retirement TSV
  72 -> 60 fields; session hotspot production tightened 73601 -> 73509

New focused tests, positive and negative on each cut:

- `inspect_snapshot_reads_honor_level_from_the_canonical_player_like_cpp`
- `inspect_snapshot_sees_a_canonical_honor_change_without_a_registry_sync_like_cpp`
- `inspect_snapshot_reports_zero_honor_without_a_canonical_owner_like_cpp`
- `collect_legacy_creature_aggro_candidates_reads_reputation_and_flags_from_canonical_like_cpp`
- `collect_legacy_creature_aggro_candidates_default_reputation_and_flags_without_canonical_like_cpp`

## Two things worth a reviewer's attention

The quest-share harness was mirroring production incorrectly:
`install_represented_party` registered a receiver in the directory but never put
a canonical `Player` on a map. The cut surfaced it, and the harness now installs
one the way a live session does. Loosening the assertion instead would have
hidden the invariant the cut depends on.

One quest test wrote the receiver's reputation twice, before and after the
registry sync — the second write existed only because the republish overwrote the
first. That duplicate is gone.

Refs #133, #252, #297.


alseif0x:
Superseded by #377 with identical commits.

This branch was created with `gh issue develop 252`, which gives it a development link to the issue. GitHub reported the PR as closing #252 even after the body carried no closing keyword, and #252 must stay open: its remaining ~53 fields are session-owned and blocked on Player state leaving `WorldSession` (see the measurement on the issue). Reopening it after an accidental merge-close is not the same as never closing it, so the commits moved to an unlinked branch. #377 reports zero closing references.

